PR #2, opened 2022-12-08
Consolidate Dependabot Change Note Workflow

rmartin16:
## Dependabot Change Note
I sourced `dependabot-changenote.yml` from `rubicon-objc` since it includes the `printf` fix.

This turned out even easier than I expected. The workflow below is all that'll be necessary for a repo to add dependabot change notes (once this PR is merged).

```yaml
name: Dependabot Change Note

on:
  push:
    branches:
      - 'dependabot/**'

jobs:
  changenote:
    name: Dependabot Change Note
    uses: beeware/.github/.github/workflows/dependabot-changenote.yml@main
    secrets: inherit
```

Example PR: https://github.com/rmartin16/briefcase-test-1/pull/6
Change note workflow run: https://github.com/rmartin16/briefcase-test-1/actions/runs/3653055359

## PR Checklist:

- [x] All new features have been tested
- [x] All new features have been documented
- [x] I have read the **CONTRIBUTING.md** file
- [x] I will abide by the code of conduct

PR #3, opened 2022-12-11
Enable dependabot on the dependabot configuration

freakboy3742:
The shared Dependabot CI workflow added by #2 will also need to be updated by dependabot.

There's a chicken-and-egg problem here; this repo defines a dependabot workflow to add change notes; which means there is a workflow that needs tone updated by dependabot. However, this repo *doesn't* use towncrier (since there's nothing to release). To allow for this, an empty placeholder directory has been added where change notes can be written (and then ignored)

## PR Checklist:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates on Sunday, 8PM UTC
+      interval: "weekly"
+      day: "sunday"
+      time: "20:00"
--- a/.github/workflows/dependabot-changenote.yml
+++ b/.github/workflows/dependabot-changenote.yml
@@ -9,7 +9,7 @@ permissions:
 jobs:
   changenote:
     name: Dependabot Change Note
-    if: github.actor == 'dependabot[bot]'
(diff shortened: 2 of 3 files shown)


rmartin16 (commented):
@freakboy3742,
I was already considering this....even more so now given the dependabot changenote worklfow doesn't need to run for this repo.

It may make more sense to move the dependabot changenote workflow to a different directory such as `shared_workflows` or the like.

I think that'd prevent it from automatically running in this repo and make it more clear why the workflow is located in this repo at all.

Thoughts?

freakboy3742:
@rmartin16 If the "shared workflow" folder approach will work, that definitely makes sense to me.

rmartin16:
whelp....

https://github.com/orgs/community/discussions/9050

Tried it just to confirm in my test setup:
image

guess we're stuck with this for now

rmartin16 (commented):
Ok...another approach since we apparently can't move the workflow.

We could add a condition to only run the workflow for other repos.

In `dependabot-changenote.yml`, update:
```
if: github.actor == 'dependabot[bot]'
```

to:
```
if: github.actor == 'dependabot[bot]' && github.repository != 'beeware/.github'
```

freakboy3742:
That makes sense. Other repos have to opt-in to the change note behavior by importing this workflow explicitly; so adding a self-exclusion should be sufficient AFAICT.

rmartin16 (approved):

PR #5, opened 2023-01-25
Add composite workflow action to install python from the Windows Store

rmartin16:
## Changes
To allow for better testing of beeware/briefcase#1054 and the impacts of Windows Store Python in general, this introduces a github action workflow to install Python from the Window Store. The aspiration is to support replacing any use of `actions/setup-python`.

This is a [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) to allow it to run as a step in a predefined job. As opposed to a workflow that runs as an isolated job.

## Notes
Currently supports Python 3.8 to Python 3.11. There is a disclaimer in the store for Python 3.7 saying its "primarily for evaluation purposes, and not all features are guaranteed stable".

Notably, though, Python 3.8 and 3.9 don't support running `tox` apparently (or at least the way we're running it); the issue seems to boil down to access to the python exe; shown in CI [here](https://github.com/beeware/briefcase/actions/runs/4008261600/jobs/6882366804) and in a local VM below:
```
❯ python
Python 3.9.13 (tags/v3.9.13:6de2ca5, May 17 2022, 16:36:42) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> from pathlib import Path
>>> Path(sys.executable).resolve()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.9_3.9.3568.0_x64__qbz5n2kfra8p0\lib\pathlib.py", line 1215, in resolve
    s = self._flavour.resolve(self, strict=strict)
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.9_3.9.3568.0_x64__qbz5n2kfra8p0\lib\pathlib.py", line 215, in resolve
    s = self._ext_to_normal(_getfinalpathname(s))
OSError: [WinError 1920] The file cannot be accessed by the system: 'C:\\Users\\user\\AppData\\Local\\Microsoft\\WindowsApps\\PythonSoftwareFoundation.Python.3.9_qbz5n2kfra8p0\\python.exe'
```

Additionally, the use of https://store.rg-adguard.net/ is a bit suspect...but this appears to be the primary method to obtain direct links to download packages from Windows Store. While this source is certainly initially dubious, the ultimate download link provided is from microsoft (as seen [here](https://github.com/beeware/briefcase/actions/runs/4002352123/jobs/6869509337)).

## PR Checklist:

### DIFF
--- a/.github/actions/install-win-store-python/action.yml
+++ b/.github/actions/install-win-store-python/action.yml
@@ -11,40 +11,69 @@ runs:
   using: composite
   steps:
 
-    - name: Python Constants
+    - name: Python Definitions
       id: python
       shell: powershell
       run: |
-        # Constants for all Python versions
-        echo "exe-dir=$(Join-Path -Path $env:LOCALAPPDATA -ChildPath \Microsoft\WindowsApps\)" | Out-File -FilePath $Env:GITHUB_OUTPUT -Encoding utf8 -Append
-
-        # Constants for each Python version
-        if ( "${{ inputs.python-version }}" -eq "3.11" ) {
-          echo "version=3.11" | Out-File -FilePath $Env:GITHUB_OUTPUT -Encoding utf8 -Append
-          echo "store-url=https://apps.microsoft.com/store/detail/python-311/9NRWMJP3717K" | Out-File -FilePath $Env:GITHUB_OUTPUT -Encoding utf8 -Append
-          echo "scripts-dir=$(Join-Path -Path $env:LOCALAPPDATA -ChildPath '\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\local-packages\Python311\Scripts')" | Out-File -FilePath $Env:GITHUB_OUTPUT -Encoding utf8 -Append
+        switch ( "${{ inputs.python-version }}" ) {
+          "3.11" {
+            $StoreURL = "https://apps.microsoft.com/store/detail/python-311/9NRWMJP3717K"
+            $ExeDir = Join-Path "$env:LOCALAPPDATA" "\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0"
+            $ScriptsDir = Join-Path "$env:LOCALAPPDATA" "\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\local-packages\Python311\Scripts"
+          }
+          "3.10" {
+            $StoreURL = "https://apps.microsoft.com/store/detail/python-310/9PJPW5LDXLZ5"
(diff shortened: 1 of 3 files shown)


rmartin16:
Demonstrated here: https://github.com/beeware/briefcase/actions/runs/4002844003/jobs/6870449742

.github/actions/install-win-store-python/action.yml
```diff
@@ -0,0 +1,51 @@
+name: Install Windows Store Python
+description: Installs Python from the Windows Store
+
+inputs:
+  python-version:
+    description: 'Python version to install'
+    default: '3.11'  # should default to the latest Python
```
freakboy3742, line 7: "3.X" will give you the latest stable release (3.11.1 at present)

.github/actions/install-win-store-python/action.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    - name: Python Constants
+      id: python
+      shell: powershell
+      run: |
+        # Constants for all Python versions
+        echo "exe-dir=$(Join-Path -Path $env:LOCALAPPDATA -ChildPath \Microsoft\WindowsApps\)" | Out-File -FilePath $Env:GITHUB_OUTPUT -Encoding utf8 -Append
+
+        # Constants for each Python version
+        if ( "${{ inputs.python-version }}" -eq "3.11" ) {
```
freakboy3742, line 22: Presumably we could/should fill this out for every other version we want to support, providing a download URL?

freakboy3742 (changes requested):
This looks good to me (complex, for Windows-related reasons, but good) - nice work! 

If we're going to commit this to the core repo, do we want to fill out the "versions other than 3.11" section?

.github/actions/install-win-store-python/action.yml
rmartin16: Added back to Python 3.8. Added notes about the....usability of these Pythons before 3.10...

.github/actions/install-win-store-python/action.yml
rmartin16: I'm not sure I'm following...given we need to manually specify the specific python we want somewhere, using `3.X` here would just require a translation later, right?

rmartin16:
I added some basic tests to confirm the install finishes and the default python is at the expected place and is the expected version. Feedback welcome.

.github/actions/install-win-store-python/action.yml
freakboy3742: My mistake - I read this as a question about "how can we default to the latest python". I guess it might be nice to support "3.X" syntax, but I wouldn't consider that a blocker by any means.

freakboy3742 (approved):
Nice work - and the tests are especially appreciated.

PR #8, opened 2023-02-04
Centralize More Universal CI Actions

rmartin16:
### `pre-commit-run`
- Runs `pre-commit` checks for an arbitrary repo.
- `pre-commit-source` defaults to `.[dev]`; use `pre-commit` if the repo doesn't define a pinned version.
- Example usage:
```
jobs:
  pre-commit:
    name: Pre-commit checks
    uses: beeware/.github/.github/workflows/pre-commit-run.yml@main
    with:
      pre-commit-source: './core[dev]'
```

### `towncrier-run`
- Runs `tox -e towncrier-check` for an arbitrary repo.
- `tox-source` defaults to `.[dev]`; use `tox` if the repo doesn't define a pinned version.
- Example usage:
```
jobs:
  towncrier:
    name: Check towncrier
    uses: beeware/.github/.github/workflows/towncrier-run.yml@main
    with:
      tox-source: './core[dev]'
```

### `package-python`
- Creates a Python package via `tox -e package` and uploads as an artifact named 'packages-\<repo name\>'.
- Example usage:
```
jobs:
  package:
    name: Python Package
    uses: beeware/.github/.github/workflows/package-python.yml@main
    with:
      tox-source: './core[dev]'
      build-subdirectory: ${{ matrix.subdir }}
      distribution-path: '*/dist/*'
    strategy:
      matrix:
        subdir:
        - 'android'
        - 'cocoa'
        ...
```

### `verify-app-build`
- Builds an app for a target platform, format, and framework on a particular OS using the invoking repo.
- `runner-os` and `framework` are required at minimum.
- If `target-platform` and `target-format` are not explicitly define, then all apps compatible with the `runner-os` and `framework` are built.
  - Example usage:
```
jobs:
  verify-apps:
    name: Build apps
    uses: beeware/.github/.github/workflows/verify-app-build.yml@main
    with:
      python-version: ${{ matrix.python-version }}
      python-source: ${{ matrix.python-source }}
      runner-os: windows-latest
      framework: ${{ matrix.framework }}
      target-platform: windows
      target-format: VisualStudio
    strategy:
      fail-fast: false
      matrix:
        python-version: [ '3.8', '3.9', '3.10', '3.11' ]
        python-source: [ 'github', 'winstore' ]
        framework: ['toga', 'pyside2', 'pyside6', 'ppb']
```

### `install-briefcase` 
- In several CI runs, we need to install Briefcase; however, the appropriate version to install is dependent on the context.
- When CI runs on the `main` branch in a repo, Briefcase should be installed from its `main` branch.
  - Example: https://github.com/rmartin16/briefcase-windows-app-template-test-1/actions/runs/4086765608/jobs/7049375420
- When CI runs for an arbitrary branch, Briefcase should be installed using branch as version.
  - Example for `v0.3.11` branch: https://github.com/rmartin16/briefcase-windows-app-template-test-1/actions/runs/4087055913/jobs/7049392731
  - Example for `v10.0.0` branch: https://github.com/rmartin16/briefcase-windows-app-template-test-1/actions/runs/4087072170/jobs/7049409704
- When CI runs for a tag, Briefcase should be installed using the tag version.
  - Example for `v40.10.6` tag: https://github.com/rmartin16/briefcase-windows-app-template-test-1/actions/runs/4087695409/jobs/7049424029
  - Example for `v0.3.12` tag: https://github.com/rmartin16/briefcase-windows-app-template-test-1/actions/runs/4087711584/jobs/7049438896
- When CI runs for a PR, Briefcase should be installed using PR's target branch version.
  - For instance, when a PR for a template is targeting the `v0.3.12` branch, the `v0.3.12` version of Briefcase should be installed.
  - Example for `main` branch: https://github.com/rmartin16/briefcase-windows-app-template-test-1/actions/runs/4086946506/jobs/7049469167
    - Of note, we are _really_ close to the maximum windows path length (hence the build failures) 😟 
  - Example for `v0.3.11` branch: https://github.com/rmartin16/briefcase-windows-app-template-test-1/actions/runs/4086976986/jobs/7049493706
  - Example for `v10.0.0` branch: https://github.com/rmartin16/briefcase-windows-app-template-test-1/actions/runs/4087022947/jobs/7049517402
    - Briefcase's `main` branch is used when the version is unknown.
- Example usage:
```
jobs:
  job-name:
  runs-on: ubuntu-latest
  steps:
    - name: Install Briefcase
      uses: beeware/.github/.github/actions/install-briefcase@main
```

### `ci` worklfow for `.github`
- Exercises each of the workflows in the context of where they will be used.
- In practice, this should help detect issues with changes to these workflows before those issues surface in the downstream repos.
- This also means additional repos should be added here when such repos implement these workflows.

## Implementation PRs
- beeware/briefcase#1080
- beeware/toga#1771
- beeware/rubicon-objc#269
- beeware/Python-support-testbed#17
- beeware/gbulb#68
- beeware/briefcase-windows-app-template#5
- beeware/briefcase-windows-VisualStudio-template#6
- beeware/briefcase-iOS-Xcode-template#16
- beeware/briefcase-android-gradle-template#63
- beeware/briefcase-web-static-template#4
- beeware/briefcase-linux-appimage-template#25
- beeware/briefcase-linux-flatpak-template#11
- beeware/briefcase-macOS-app-template#13
- beeware/briefcase-macOS-Xcode-template#21

## PR Checklist:


rmartin16:
@freakboy3742, this should be ready for review now. I've got it working everywhere I think we need to use it.

.github/workflows/pre-commit-run.yml
```diff
@@ -0,0 +1,58 @@
...
+      uses: actions/cache@v3.2.4
+      with:
+        path: ~/.cache/pre-commit
+        key: pre-commit-3|${{ env.pythonLocation }}|${{ hashFiles('.pre-commit-config.yaml') }}
+
+    - name: Install pre-commit
+      run: python -m pip install ${{ inputs.pre-commit-source }}
+
+    - name: Run pre-commit
+      run: pre-commit run --show-diff-on-failure --color=always --all-files
```
freakboy3742, line 58: Why a manual install and execution, rather than the action provided by pre-commit itself?

.github/workflows/towncrier-run.yml
```diff
@@ -0,0 +1,62 @@
...
+      with:
+        repository: ${{ inputs.repository }}
+        fetch-depth: 0
+
+    - name: Setup Python
+      uses: actions/setup-python@v4.5.0
+      with:
+        python-version: ${{ inputs.python-version }}
+        cache: 'pip'
+        cache-dependency-path: '**/setup.cfg'
```
freakboy3742, line 56: I wonder if we might want to add `pyproject.toml` and `setup.py` to this list as well. We're still using `setup.py` in toga; I imagine that a move to full `pyproject.toml` format in the future is inevitable (either as part of a setuptools upgrade, or a move to hatch), so including those files would seem like a good protective mechanism.

.github/workflows/package-python.yml
```diff
@@ -0,0 +1,80 @@
...
+        with:
+          repository: ${{ inputs.repository }}
+          fetch-depth: 0  # Fetch all refs so setuptools_scm can generate the correct version number
+
+      - name: Setup Python
+        uses: actions/setup-python@v4.5.0
+        with:
+          python-version: ${{ inputs.python-version }}
+          cache: 'pip'
+          cache-dependency-path: '**/setup.cfg'
```
freakboy3742, line 62: As above - do we want to add `pyproject.toml` and `setup.py` to this list?

.github/workflows/package-python.yml
```diff
@@ -0,0 +1,80 @@
+name: Create Python Package
```
freakboy3742, line 1: The "package-python" name is a little misleading... at the very least, it's "Python-package"; but if it's a beeware project, it's going to be Python regardless.

.github/actions/install-briefcase/action.yml
```diff
@@ -0,0 +1,101 @@
+name: Install Briefcase
+description: Installs a version of Briefcase appropriate for the context.
+
+inputs:
+  briefcase-url:
+    description: 'URL to Briefcase git repository.'
```
freakboy3742, line 6: We've been trying to converge on " rather than ' in config files (for broad consistency with black); is there a particular reason you've reverted a bunch of those changes and moved to '?

.github/actions/install-briefcase/action.yml
```diff
@@ -0,0 +1,101 @@
...
+        
+        # Default to Briefcase's default branch
+        else
+          echo "::warning::Unknown Briefcase version: '${TARGET_VERSION}'; defaulting to '${{ inputs.briefcase-default-branch }}'."
+          echo "version=${{ inputs.briefcase-default-branch }}" >> ${GITHUB_OUTPUT}
+        fi
+
+    - name: Install Briefcase
+      shell: bash
+      run: python -m pip install -U git+${{ inputs.briefcase-url }}@${{ steps.briefcase.outputs.version }}
```
freakboy3742, line 101: Confirming my understanding - this means that *all* briefcase installs will be running against tags/branches from github - there's no case where we'll be using a published wheel? I guess there's a technical edge case if the briefcase tag doesn't match the exact code used to release the wheel... this *shouldn't* happen, but it's an interesting potential discrepancy.

.github/workflows/verify-app-build.yml
```diff
@@ -0,0 +1,255 @@
...
+      with:
+        name: packages-${{ steps.package.outputs.name}}
+        path: dist
+
+    - name: Install Briefcase artifact
+      if: endsWith(inputs.repository, 'briefcase')
+      run: python -m pip install dist/briefcase-*.whl
+      
+    - name: Install Briefcase
+      if: endsWith(inputs.repository, 'briefcase') != true
```
freakboy3742, line 129: Is this not just `!endWith...`?

.github/workflows/verify-app-build.yml
```diff
@@ -0,0 +1,255 @@
...
+        default: ${{ github.repository }}
+        type: string
+      target-platform:
+        description: 'The target platform for the app, i.e. linux, macos, or windows. Leave blank all supported platforms.'
+        default: ''
+        type: string
+      target-format:
+        description: 'The target format for the app, e.g. appimage, xcode, app, etc. Leave blank all supported formats.'
+        default: ''
+        type: string
```
freakboy3742, line 38: To clarify my understanding: there wouldn't be a way to specify "macOS app and Windows VisualStudio", because "target-platform" would be "macos, windows" and target-format would be "app, VisualStudio", which would then catch "Windows app" - correct? I don't think this is a dealbreaker - in practice, we're either doing everything, or doing one specific platform - but it's worth being explicit. The only real alternative I can think of would to be make target-platform a single string, rather than a list; (so - target-platform, but target-format*s*), and then do an exact match rather than a contains when checking target-platform.

.github/actions/install-briefcase/action.yml
```diff
@@ -0,0 +1,101 @@
+name: Install Briefcase
+description: Installs a version of Briefcase appropriate for the context.
```
freakboy3742, line 2: Is there any particular reason for the `actions/install-briefcase/action.yml` structure? Could this be simplified to `actions/install-briefcase.yml`?

freakboy3742 (changes requested):
Woof... that's a lot of changes and dependent PRs :-)

I've flagged a few questions and comments inline; many of them are clarifications for my own benefit to make sure I fully understand what is going on; but broadly speaking, this is going in the right direction (and should make our lives a whole lot easier... once we land the changes 😅).

I haven't fully reviewed all the downstream repos yet - I've been working on the theory that we need to get this one right first, then get the other repos to be consistent with whatever lands. So - my comments here are mostly about the API surface and operation, guided by what I can see as the pre-existing CI usage, and intended usage once this PR is in place. 

That said - a couple of things that I noticed:

1. In the downstream repos (esp. Briefcase and Toga), I can see you've updated the CI target, but not the package target; that will be pulling artefacts from the CI build, so it will be impacted by the name change on the package workflow.
2. I can't see anywhere in the Briefcase config that hits the Windows store python builds - am I missing something?

.github/workflows/pre-commit-run.yml
rmartin16: Just so we use the version of `pre-commit` that's pinned for the repo and CI doesn't start reporting errors that wouldn't be reported locally due to versions differences.

I'm mostly ambivalent (given the likelihood of issue is quite low), though, if you'd rather continue using the action from `pre-commit`.

.github/actions/install-briefcase/action.yml
rmartin16: Right; all Briefcase installs using this action would come from the GitHub repo and never from PyPI.

.github/workflows/verify-app-build.yml
rmartin16: Currently, this script does not support multiple values for `target-platform` or `target-format` (or any of the other inputs for that matter).

This script is a combination of the use-cases from Briefcase and the template repos. Briefcase builds as many Apps as are possible for a framework and OS combo. The template repos only build a specific platform/format at any one time.

So, to satisfy these two use-cases, you can specify a specific target platform/format and build that while also allowing a simpler matrix based only on OS and framework while building all platform/format combos.

So, at the end of the day, this workflow is designed to to be used in a matrix that minimally defines an OS and a framework.

.github/actions/install-briefcase/action.yml
rmartin16: This is a requirement [defined](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions) by GitHub for composite actions.

.github/workflows/verify-app-build.yml
rmartin16: definitely is....my brain didn't want to believe `!` was supported or something...

Updated.

.github/workflows/verify-app-build.yml
freakboy3742: Got it - I misread/misunderstood how the `fromJSON` matching was working.

.github/actions/install-briefcase/action.yml
freakboy3742: TIL.

.github/workflows/package-python.yml
rmartin16: So....this starts to get a little tricky. The `setup-python` action hard errors if the files specified here don't exist. I'll give it a go and see if all the repos have these files.

(On a related note, this is why some of the other uses of `setup-python` don't/can't use this built in caching.

.github/workflows/pre-commit-run.yml
freakboy3742: I guess it's a little weird to be rolling our own when an existing battle-proven action exists... but the fact that we'll be using the exact version pinned in our CI config is one less potential incompatibility between dev and test, so I can live with the duplication.

.github/workflows/package-python.yml
freakboy3742: I can save you the trouble; pyproject *should* exist everywhere (and if it doesn't, that's a bug); setup.py currently does, but *should* be slowly disappearing. (I've deleted it in the deb PR branch, for example).

.github/workflows/pre-commit-run.yml
rmartin16: FWIW, I literally just copied `pre-commit`'s current [implementation](https://github.com/pre-commit/action/blob/main/action.yml).

.github/workflows/package-python.yml
freakboy3742: Oh - and the template repositories won't have any of these files - but then, we won't be building Python packages for those repos.

.github/workflows/package-python.yml
rmartin16: That's definitely fair.

With that, GitHub currently has a limitation requiring all workflows to exist in the `.github/workflows/` directory. So, while we may not end up with too many workflows, I was looking for a way for workflows related to the same thing/tool/concept to alphabetically. That's why I started the workflows for `pre-commit` with `pre-commit` and same with `towncrier`. This leads, though, to a bit of awkward grammar with the action word coming last.

With that in mind, perhaps `package-beeware-create.yml` best follows those rules (although equally awkward)....and perhaps where a workflow like `package-beeware-deploy.yml` could theoretically exist alongside it...

.github/actions/install-briefcase/action.yml
rmartin16: Mostly because `"` and `'` have different functional meanings in YAML....while in Python they are functionally equivalent.

Double quoted strings in YAML pass through a level of escape parsing that single quotes strings do not. So, really, I'm trying to emphasize that I literally want the value between the two single quotes without any interpretation.

That said, none of this is particularly consistent. For instance, all the places that GitHub is substituting values for `${{ github.event_name }}` (and the like), practically none of those are quoted at all.

So, unless this is compelling, I'm fine converging on double quotes. I also just realized that I was confusing TOML with YAML; TOML is a bit more restrictive about double quotes...

.github/workflows/verify-app-build.yml
rmartin16: interestingly....GitHub allows `endsWith(inputs.repository, 'briefcase') != true` but requires the brackets for `${{ !endsWith(inputs.repository, 'briefcase') != true }}`....hmm...

.github/workflows/package-python.yml
rmartin16: Landed on `python-package-create.yml`.

rmartin16:
> I haven't fully reviewed all the downstream repos yet - I've been working on the theory that we need to get this one right first, then get the other repos to be consistent with whatever lands. So - my comments here are mostly about the API surface and operation, guided by what I can see as the pre-existing CI usage, and intended usage once this PR is in place.

This is definitely the approach I intend.

> 
> That said - a couple of things that I noticed:
> 
> 1. In the downstream repos (esp. Briefcase and Toga), I can see you've updated the CI target, but not the package target; that will be pulling artefacts from the CI build, so it will be impacted by the name change on the package workflow.

I think you're talking about, e.g. the `release.yml` in Briefcase where it download the uploaded package. That's definitely a good point. This is one of my weaker areas in understanding about the CI processes.

Reviewing `release.yml`, it runs `ci.yml` and then creates a release with the produced artifact. Based on that, I think I'll update Briefcase's `ci.yml` to propagate the `artifact-name` output from `python-package-create.yml` so `release.yml` can use it.

> 2. I can't see anywhere in the Briefcase config that hits the Windows store python builds - am I missing something?

There's still the extra `include` to run Briefcase unit tests against one version of Windows Store Python:
```
        # Run tests against the latest Windows Store Python
        - platform: "windows"
          python-version: "winstore"
          experimental: false
```

.github/workflows/package-python.yml
rmartin16: The tests finished after I added `pyproject.toml` alongside `setup.cfg`.

.github/workflows/towncrier-run.yml
rmartin16: Added `**/pyproject.toml` as a path.

rmartin16:
I removed the testing code last night to allow `app-build-verify.yml` to work from the branch from this PR. So, I'll need to wait until this is merged to finish updating the dependent PRs.

```yaml
    - name: Checkout beeware/.github
      uses: actions/checkout@v3.3.0
      with:
        repository: "beeware/.github"
        path: "beeware-.github"
```

freakboy3742 (approved):
I picked up a couple of minor whitespace issues which I've pushed; but otherwise, I think this is good to go.

rmartin16:
ohh. interesting....well, I've added `.pre-commit-config.yaml` to the other PRs....probably should have done that here as well.

freakboy3742:
I haven't merged yet - so if there's something else you want to squeeze in here, I can hold off...

rmartin16:
All right. Added the same basic `.pre-commit-config.yaml` I put in the other repos.

PR #10, opened 2023-02-12
Add a fallback for cron invocation of pre-commit-update trigger.

freakboy3742:
When pre-commit-update is triggered on a cron, it doesn't have a `pre-commit-source` input, so it fails because the input evaluates as "". This falls back to `pre-commit` when there isn't a value. Calls via `workflow-call` have `.[dev]` as a default.

## PR Checklist:


rmartin16 (approved):
Manual run succeeded.

https://github.com/beeware/.github/actions/runs/4158889423/jobs/7194419219

PR #17, opened 2023-03-01
Use `perl` for regex for `install-briefcase`

rmartin16:
Since `grep -P` is not valid for BSD grep (and whatever is installed on Windows), updated to just use `perl` directly.

## PR Checklist:

### DIFF
--- a/.github/actions/install-briefcase/action.yml
+++ b/.github/actions/install-briefcase/action.yml
@@ -36,16 +36,18 @@ runs:
       id: briefcase-override
       shell: bash
       run: |
-        set +o pipefail  # prevent grep failing the step if repo/ref isn't found
-
         # Source version from Body of PR
         PR_BODY="${{ inputs.testing-pr-body }}"
         if [ -z "${PR_BODY}" ]; then
           API_URL="${{ github.api_url }}/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -24,11 +24,12 @@ jobs:
   test-install-briefcase:
     name: Test Install Briefcase
     needs: pre-commit
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner-os }}
     timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
+        runner-os: [ "ubuntu-latest", "windows-latest", "macos-latest" ]
```
freakboy3742, line 32: Thorough is good... but this repo runs over 80 configurations *before* this PR. Do we have to run every possible combination on every operating system?

rmartin16: the limit is 256 i think :) but yeah....i'll take it back out. the tests did pass for all OSes though.

rmartin16:
For posterity....`perl` definitely doesn't seem like the best solution here....however, my regex skills limit my ability to write this in native bash via `[[ "${PR_BODY}" =~ <REGEX> ]] && REF=${BASH_REMATCH[0]}`.

freakboy3742 (approved):
Looks good. 

I agree it would be nice to avoid perl... but it's available, and it works, so I guess there's no harm.

PR #18, opened 2023-03-01
Use authentication when hitting the Github API.

freakboy3742:
We've started hitting the API rate limit when querying the Github API. e.g.,

https://github.com/beeware/briefcase-web-static-template/actions/runs/4273741637/jobs/7496132787

We should be able to get a vastly increased rate limit by using authenticated API access.

## PR Checklist:


rmartin16:
This'll probably require `secrets: inherit` everywhere these actions/workflows are used....that's my guess anyway for the failures

freakboy3742:
This will also need an update to all the template repos to pass in a token with read permissions; plus an update to briefcase repo with the same.

rmartin16 (commented):
I implemented a slightly different approach in #19. Let me know what you think.

freakboy3742:
Yours is better :-)

PR #19, opened 2023-03-01
Use Authenticated GitHub API calls via `gh` and `github.token`

rmartin16:
## Summary
This is an alternative implementation for #18 to avoid API calls being rate limited.

As we've discovered, GitHub aggressively rate limits unauthenticated API calls. To authenticate, the GitHub API needs a valid token in the HTTP.

## Changes
- Use GitHub's `gh` tool to talk to the API
  - This is more robust than `curl`ing and I probably should have used it from the get go.
- Use the implicitly provided GitHub token for the workflow run
  - For each workflow run, GitHub generates a token that's valid for the length of the run.
  - The token is exposed via the `github` context as `github.token`.
  - When a workflow invokes an action or workflow, the `github` context is [implicitly passed](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow) to the called workflow.
  - The `gh` tool will use the `GITHUB_TOKEN` env var to authenticate any API calls.
  - Permissions for `github.token`
    - The [permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) for this auto-generated token can be controlled at many levels from the Organization down to inside the yaml.
    - This includes a `read` permission for `pull_request`.
    - However, I can only reason that this permission is necessary for private repos. I mean, the API endpoints for pull requests in public repos are available to the whole world via unauthenticated requests.
    - Therefore, I do not expect calling workflows to need to explicitly set this permission.

## Examples
- Dependabot changenote [workflow run](https://github.com/rmartin16/briefcase-windows-app-template-test-1/actions/runs/4305721605/jobs/7508565863)
- PR without Briefcase Repo or Ref [workflow run](https://github.com/rmartin16/briefcase-windows-app-template-test-1/actions/runs/4305721734/jobs/7508565763)
- PR with Briefcase Repo or Ref [workflow run](https://github.com/rmartin16/briefcase-windows-app-template-test-1/actions/runs/4305766600/jobs/7508677431)

## PR Checklist:

### DIFF
--- a/.github/actions/install-briefcase/action.yml
+++ b/.github/actions/install-briefcase/action.yml
@@ -41,7 +41,7 @@ runs:
         # Source version from Body of PR
         PR_BODY="${{ inputs.testing-pr-body }}"
         if [ -z "${PR_BODY}" ]; then
-          PR_BODY=$(gh api repos/{owner}/{repo}/pulls/${{ github.event.pull_request.number }} --jq '.body')
+          PR_BODY=$(gh api repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }} --jq '.body')
           printf "::group::Retrieved PR Body\n${PR_BODY}\n::endgroup::\n"
         fi
         BRIEFCASE_REPO=$(printf "%s" "${PR_BODY}" | perl -wlne '/BRIEFCASE[-_]*REPO:\s*\K\S+/i and print $&')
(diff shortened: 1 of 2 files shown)


rmartin16:
inb4: one may notice I use `${{ github.event.pull_request.number }}` in `install-briefcase` but then use `gh api` in the `dependabot-changenote` to retrieve the PR number. This is necessary because `dependabot-changenote` is designed to be triggered for `push` events; therefore, `github.event.pull_request` will not exist. The history for why it uses `push` is in beeware/briefcase#885.

freakboy3742 (approved):
Well that's a *lot* more elegant. Nice work.

PR #20, opened 2023-03-01
Don't try to retrieve the PR number if this run isn't due to a PR.

freakboy3742:
#19 worked great in CI, but failed when the PR was merged because it relied on the existence of a PR number during the CI run. 

Side note: This does mean that PRs that pass because of a `B***case-repo:` (redacted so it doesn't catch on *this* pr... 🤦 ) configuration will fail when the PR is merged. I'm not sure there's much we can do about that, though - this is just part of the "chicken and egg" problem.

## PR Checklist:

### DIFF
--- a/.github/actions/install-briefcase/action.yml
+++ b/.github/actions/install-briefcase/action.yml
@@ -38,9 +38,10 @@ runs:
         GITHUB_TOKEN: ${{ github.token }}
       shell: bash
       run: |
-        # Source version from Body of PR
+        # Source version from Body of PR (if there is one; if there isn't,
+        # github.event.pull_request.number will be empty)
         PR_BODY="${{ inputs.testing-pr-body }}"
-        if [ -z "${PR_BODY}" ]; then
+        if [ -z "${PR_BODY}" -a "${{ github.event.pull_request.number }}" ]; then
           PR_BODY=$(gh api repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }} --jq '.body')
           printf "::group::Retrieved PR Body\n${PR_BODY}\n::endgroup::\n"
         fi
--- a/.github/actions/install-briefcase/action.yml
+++ b/.github/actions/install-briefcase/action.yml
@@ -41,7 +41,7 @@ runs:
         # Source version from Body of PR (if there is one; if there isn't,
         # github.event.pull_request.number will be empty)
         PR_BODY="${{ inputs.testing-pr-body }}"
-        if [ -z "${PR_BODY}" -a "${{ github.event.pull_request.number }}" ]; then
+        if [[ -z "${PR_BODY}" && -n "${{ github.event.pull_request.number }}" ]]; then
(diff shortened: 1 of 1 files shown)


rmartin16 (approved):
Using `-a` makes me a little nervous; I think its use is usually not recommended.

For instance, in `man test`:
```
       NOTE: Binary -a and -o are inherently ambiguous.  Use 'test EXPR1 && test EXPR2' or 'test EXPR1 || test EXPR2' instead.
```

Nonetheless, we're likely to avoid its known pitfalls here; I'll leave it to you whether we should use `if [ -z "${PR_BODY}" ] && [ -n "${{ github.event.pull_request.number }}" ]` or `if [[ -z "${PR_BODY}" && -n "${{ github.event.pull_request.number }}" ]]` or similar instead.

rmartin16:
> Side note: This does mean that PRs that pass because of a `B***case-repo:` (redacted so it doesn't catch on _this_ pr... facepalm ) configuration will fail when the PR is merged. I'm not sure there's much we can do about that, though - this is just part of the "chicken and egg" problem.

Related to this....I think that GitHub recently released their implementation of a merge queue. If my understanding is correct, it will only run CI on the merge branch and not repeat it on the `main` branch. While heavy handed perhaps in our case....it may remove this problem at least.

freakboy3742:
> ```
> if [[ -z "${PR_BODY}" && -n "${{ github.event.pull_request.number }}" ]]
> ```
> 
> or similar instead.

Good call - old UNIX habits die hard :-)

As for the merge queue changes - I saw those get announced, but it wasn't immediately clear to me that it would help us. My read was that it would minimize the number of builds if you were landing multiple PRs at the same time, but in the simple case of merging a single PR, it wouldn't help much. The merge queue might help a little in the "Monday morning Dependabot backlog" case - but that's not enough of a problem that I've been motivated to investigate further. 

If you feel like investigating this more, I'm happy to incorporate any suggestions you have.

rmartin16:
> As for the merge queue changes - I saw those get announced, but it wasn't immediately clear to me that it would help us.

Yeah; I'm ignoring the prevalent purpose of a merge queue and mostly focusing on one aspect. That is, since a merge queue seems to require that a PR's CI must pass while also incorporating all commits in the target branch, it will not re-run CI once the PR is merged in to the target branch (....since it effectively already ran those tests). Although, it isn't entirely clear to me if that's how it actually works...

mhsmith:
> since a merge queue seems to require that a PR's CI must pass while also incorporating all commits in the target branch, it will not re-run CI once the PR is merged in to the target branch (....since it effectively already ran those tests)

If I understand [the docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) correctly, some of this is already the case with the normal `pull_request` workflow. It doesn't run directly against the PR branch, but on a merged commit showing what would happen if the PR was to be merged at that moment. But without a merge queue, by the time the PR actually is merged, the main branch may have moved on, so it may be necessary to run CI again.

PR #21, opened 2023-03-09
Skip RPM build tests for PySide2

rmartin16:
- `fedora:37` ships with Python 3.11 and PySide2 cannot be installed.

ref: beeware/briefcase-template#59

## PR Checklist:

### DIFF
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -217,8 +217,10 @@ jobs:
         briefcase package linux system --target debian:bullseye --adhoc-sign
 
     - name: Build Linux System project (RPM, Dockerized)
+      # fedora:37 ships with Python 3.11 and PySide2 cannot be installed
       if: |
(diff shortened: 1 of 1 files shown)


freakboy3742 (commented):
I see what you're aiming at here, but I'm not sure how I feel about it. It definitely addresses the Pyside2 issue we're seeing on Fedora. However, we haven't currently got any exposure of the list of specific frameworks in `app-build-verify`; they're currently only passed in as a part of a name used in template expansion. 

If we're going to go down the customization path here, it seems to me that we should be exposing the ability for the *use* of `app-build-verify` to either (a) change the Fedora/Debian/Arch image, or (b) disable the RPM check (maybe with a `base-rpm-image=None` configuration, rather than adapting the implementation to carve out a special case for Pyside2.

freakboy3742 (approved):
Ah - now the penny drops. We need this as a workaround right now to make CI pass for all the templates. 

We probably want to look at the broader customisation problem, but for now, this gets the job done.

PR #27, opened 2023-04-02
Quote bash variables with double quotes

rmartin16:
Runs of the `pre-commit-update` workflow failed because I wrapped the bash variable for the filename in single quotes; so all the changenotes were titled `${FILENAME}`....let's use double quotes instead.

Example: https://github.com/beeware/briefcase/pull/1162

## PR Checklist:


freakboy3742 (approved):
🤦

PR #29, opened 2023-04-14
Introduce workflow to verify ReadTheDocs build

rmartin16:
## Changes
- Creates a reusable workflow that waits until RTD reports a build for a specific version is complete.
- This is intended to apply a dependency on docs builds for publishing releases.
  - See beeware/briefcase#1182

## Notes
- The intended use-case would be running this workflow when the release tag is pushed to the repo.,
- And if this workflow fails, the progression to create a GitHub release and artifacts does not occur.

## TODO
- @freakboy3742: A secret named `RTD_API_TOKEN` needs to be added to the BeeWare organization to authorize calls to the RTD API.
  - A token can be generated here: https://readthedocs.org/accounts/tokens/
- Finalize testing strategy....github workflows are so finicky...

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,27 @@ jobs:
     with:
       pre-commit-source: "pre-commit"
 
+  test-docs-build-verify:
+    name: Test Verify Docs Build
+    needs: pre-commit
+    uses: ./.github/workflows/docs-build-verify.yml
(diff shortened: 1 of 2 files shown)


freakboy3742:
@rmartin16 `RTD_API_TOKEN` has been added as an organization secret.

rmartin16:
Superseding with #30.

PR #30, opened 2023-04-15
Introduce workflow to verify ReadTheDocs build

rmartin16:
## Changes
- Creates a reusable workflow that waits until RTD reports a build for a specific version is complete.
- This is intended to apply a dependency on docs builds for publishing releases.
  - See beeware/briefcase#1182
- The BeeWare organization secret `RTD_API_TOKEN` provides token for HTTP API authorization.

## Notes
- The intended use-case would be running this workflow when the release tag is pushed to the repo.,
- And if this workflow fails, the progression to create a GitHub release and artifacts does not occur.
- Supersedes #29 
  - Created branch directly on `beeware/.github` since secrets cannot be shared across forks.

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,23 @@ jobs:
     with:
       pre-commit-source: "pre-commit"
 
+  test-docs-build-verify:
+    name: Test Verify Docs Build
+    needs: pre-commit
+    uses: ./.github/workflows/docs-build-verify.yml
+    secrets: inherit
+    with:
+      project-name: ${{ matrix.name }}
+      project-version: ${{ matrix.version }}
+    strategy:
+      matrix:
+        name: [ "briefcase", "toga" ]
+        include:
+        - name: "briefcase"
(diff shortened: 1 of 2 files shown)


rmartin16:
I can't find a good way to include a test for the [failure case](https://github.com/beeware/.github/actions/runs/4708124173/jobs/8350380103). In other places, we've used `continue-on-error` to allow a job to fail without failing the entire run; however, GitHub apparently [doesn't allow](https://github.com/beeware/.github/actions/runs/4708248177/workflow) jobs to use `continue-on-error` with `uses`...

I support I could refactor the whole workflow and perform timeout management in the bash scripting and only actually fail if not under test conditions....but that seems like a lot of overhead for something that's unlikely to break 🤞🏼

freakboy3742 (approved):
Looks good to me, and it seems to work in the "good" case; I can't think of an obvious way to test the "bad" case, so I guess we'll have to live with the uncertainty.

I've also turned on branch protection rules for the repo.

PR #31, opened 2023-04-20
Include date in winstore Python cache key to limit cache age

rmartin16:
## Changes
- While GitHub will evict caches after not being used for 7 days, the cache will exist in perpetuity as long as it is regularly referenced.
- To ensure the Windows Store Python that's installed stays up to date, the current year and month is added to the cache key so they cache is recreated each month.
- Additionally, since this the first CI run since adding the test for RTD Builds, it is now clear the test will always fail for PRs from forks so I modified the test to only run for branches on this repo.

## Notes
- Caches with the year and date in the name can be seen [here](https://github.com/beeware/.github/actions/caches).
- GitHub does provide an API to delete caches....but it seems especially over-engineered to periodically check the age of these caches and delete them manually.

## PR Checklist:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
     name: Test Verify Docs Build
     needs: pre-commit
     # inheriting secrets does not work for forked repos
-    if: ${{ github.event.pull_request.head.repo.owner == 'beeware' }}
+    if: ${{ github.event.pull_request.head.repo.owner.login == 'beeware' }}
     uses: ./.github/workflows/docs-build-verify.yml
     secrets: inherit
     with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,8 @@ jobs:
   test-docs-build-verify:
     name: Test Verify Docs Build
     needs: pre-commit
+    # inheriting secrets does not work for forked repos
+    if: ${{ github.event.pull_request.head.repo.owner == 'beeware' }}
     uses: ./.github/workflows/docs-build-verify.yml
     secrets: inherit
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -24,6 +24,8 @@ jobs:
   test-docs-build-verify:
     name: Test Verify Docs Build
     needs: pre-commit
+    # inheriting secrets does not work for forked repos
+    if: ${{ github.event.pull_request.head.repo.owner == 'beeware' }}
```
freakboy3742, line 28: Is this an intentional inclusion? It doesn't seem related to the caching problem...

freakboy3742 (changes requested):
The cache change makes sense; the extra bit about inheriting secrets doesn't seem to be related, though. I *think* it makes sense (making sure that forked repos won't try to verify the RTD build, since they won't have credentials to do so); but I wanted to make sure the inclusion was intentional.

.github/workflows/ci.yml
rmartin16: Apologies; I should have explicitly mentioned this.

It is intentional; the [first run](https://github.com/beeware/.github/actions/runs/4756507186) of CI failed because the `RTD_API_TOKEN` secret did not resolve. I considered removing the test altogether but this seemed like a better middle ground to ensure CI passes for PRs from forked repos going forward while the test can test at least run sometimes.

P.S. - when I originally added this test, I moved the PR from my fork to this repo; at the time, I was hoping that future PRs from forks wouldn't hit the same problem....I was wrong...

freakboy3742 (approved):
Ah - that makes sense - thanks for the clarification.

PR #36, opened 2023-05-06
Use `FORCE_COLOR` to enable colored output in CI

rmartin16:
## Changes
- Reverted 36cbe2fd1927c4cb2304f08b0622572077a6a2e4 in favor of setting `FORCE_COLOR=1` to activate color output in CI

## Notes
- While there isn't a standard for enabling/forcing colored output, there is some agreement about using `FORCE_COLOR`
  - For context, this all revolves around the shells in GitHub runner not [using a tty](https://github.com/actions/runner/issues/241)
  - At any rate, Python, at least, seems to have aggregated around `FORCE_COLOR` for the most part:
    - `pytest` had been using `PY_COLOR` env var for a while...but it never really caught on
    - Initially, Will was [against](https://github.com/Textualize/rich/issues/343) this in Rich; he was eventually [convinced](https://github.com/pypa/pip/issues/10909) that [respecting](https://github.com/Textualize/rich/pull/2449) `FORCE_COLOR` made sense
    - `tox` has [supported](https://github.com/tox-dev/tox/pull/1630) it since v4
    - `pytest` has [supported](https://github.com/pytest-dev/pytest/pull/7466) it since v6
    - While `pip` is supposed to respect `FORCE_COLOR`, I don't see it working in CI...
- So, I think this is a better approach than trying to activate it for each tool
  - Perhaps unfortunately, it does require setting the env var in each workflow
  - I couldn't find a way to configure GitHub to inject an env var in to every workflow run...

## PR Checklist:

### DIFF
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -45,6 +45,9 @@ defaults:
   run:
     shell: bash
 
+env:
+  FORCE_COLOR: "1"
+
 jobs:
   verify-app:
     name: Verify App Build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,13 +6,16 @@ on:
     branches:
       - main
 
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
 defaults:
   run:
     shell: bash
 
-concurrency:
-  group: ${{ github.ref }}
-  cancel-in-progress: true
+env:
+  FORCE_COLOR: "1"
 
 jobs:
   pre-commit:
--- a/.github/workflows/pre-commit-update.yml
+++ b/.github/workflows/pre-commit-update.yml
@@ -64,7 +64,7 @@ jobs:
       - name: Run pre-commit
         # pre-commit returns non-zero when files are changed and fails the job
         continue-on-error: true
-        run: pre-commit run --all-files --color=always
+        run: pre-commit run --all-files
 
       - name: PR Needed?
         id: pr
--- a/.github/workflows/pre-commit-update.yml
+++ b/.github/workflows/pre-commit-update.yml
@@ -25,6 +25,7 @@ permissions:
 env:
   BRANCH_NAME: "autoupdates/pre-commit"
   CHANGENOTE_DIR: "./changes"
+  FORCE_COLOR: "1"
 
 defaults:
   run:
--- a/.github/workflows/python-package-create.yml
+++ b/.github/workflows/python-package-create.yml
@@ -36,6 +36,9 @@ on:
         description: "Name of the uploaded artifact; use for artifact retrieval."
         value: ${{ jobs.package.outputs.artifact-name }}
 
+env:
+  FORCE_COLOR: "1"
+
 jobs:
   package:
     name: Create Python Package
--- a/.github/workflows/python-package-create.yml
+++ b/.github/workflows/python-package-create.yml
@@ -71,11 +71,11 @@ jobs:
 
       - name: Build wheels
         if: inputs.build-subdirectory == ''
-        run: tox --colored yes -e package
+        run: tox -e package
 
       - name: Build wheels from subdirectory
         if: inputs.build-subdirectory != ''
-        run: tox --colored yes -e package -- ${{ inputs.build-subdirectory }}
+        run: tox -e package -- ${{ inputs.build-subdirectory }}
 
       - name: Upload Package
         uses: actions/upload-artifact@v3.1.2
--- a/.github/workflows/towncrier-run.yml
+++ b/.github/workflows/towncrier-run.yml
@@ -64,4 +64,4 @@ jobs:
       run: python -m pip install ${{ inputs.tox-source }}
 
     - name: Run towncrier check
-      run: tox --colored yes -e towncrier-check
+      run: tox -e towncrier-check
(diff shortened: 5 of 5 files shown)


rmartin16:
hmm....this is activating the Rich progress bar....creating some definitely [undesired CI logs](https://github.com/beeware/.github/actions/runs/4902657729/jobs/8754641453).

Looking over the Rich [docs](https://rich.readthedocs.io/en/stable/console.html#environment-variables), it says:
> If the environment variable `FORCE_COLOR` is set, then color/styles will be enabled regardless of the value of `TERM`. This is useful on CI systems which aren’t terminals but can none-the-less display ANSI escape sequences.

However, immediately preceding that in the Interactive Mode [section](https://rich.readthedocs.io/en/stable/console.html#interactive-mode):
> Some CI systems support ANSI color and style but not anything that moves the cursor or selectively refreshes parts of the terminal. For these you might want to set `force_terminal` to `True` and `force_interactive` to `False`.

So, given that Rich is basically explicitly referencing this situation....and says "change the way youre initializing `Console`", I think this deserves changes in Briefcase.

Because this doesn't only affect BeeWare's own CI; this would affect anyone trying to use Briefcase in their own CI with `FORCE_COLOR`.

rmartin16:
In Briefcase, we'd have to really move some things around to control how `Console` is instantiated since its construction happens at import.

However, it is simple to disable the use of Rich animations based on an arbitrary flag. I pushed a [commit](https://github.com/rmartin16/briefcase/commit/f47d24b296259ed39cfc616279f6324d477e257d) to disable progress bars and wait bars if the `--no-input` switch is used. This produces much [better logs](https://github.com/beeware/.github/actions/runs/4902946410/jobs/8755100359) when `FORCE_COLORS` is used.

@freakboy3742, can you please provide your thoughts on these changes?
- Using `FORCE_COLORS` instead of turning colors on each tool at a time and for each invocation?
- Expanding the meaning of `--no-input` to take on more of a meaning of "non-interactive"?

freakboy3742:
> * Using `FORCE_COLORS` instead of turning colors on each tool at a time and for each invocation?

If there's an emerging environment based standard, it makes sense to me to adopt it. It's a lot easier to set this flag once, and then have every tool pick it up, rather than have to remember to turn on colors for each individual tool invocation.

However - can I get a confirmation on whether the the flag is `FORCE_COLOR` or `FORCE_COLORS`? The implementations all seem to say `FORCE_COLOR`, but a lot of the descriptions here say `FORCE_COLORS`... I just want to make sure this is a typo, and not some deep seated schism in the community that we're diving into :-)

> * Expanding the meaning of `--no-input` to take on more of a meaning of "non-interactive"?

Definitely agree that there's a need to be able to disable the animations; I guess my question is whether we should overload `--no-input`, or use a more traditional TTY detection approach. I'm guessing FORCE_COLOR is overriding default TTY detection, which is why it's made a mess of the iOS logs; not sure if we can pull back the "color, but not TTY" aspect easily. 

If we can't - the only practical use of `--no-input` is in a CI configuration, so I don't fundamentally object to `--no-input` being overloaded.

rmartin16:
> However - can I get a confirmation on whether the the flag is `FORCE_COLOR` or `FORCE_COLORS`?

It's `FORCE_COLOR`; definitely my typo. However, there is definitely discontent with the fact that `FORCE_COLOR=0` disables color in some apps....and enables it in others....

> > * Expanding the meaning of `--no-input` to take on more of a meaning of "non-interactive"?
> 
> Definitely agree that there's a need to be able to disable the animations; I guess my question is whether we should overload `--no-input`, or use a more traditional TTY detection approach. I'm guessing FORCE_COLOR is overriding default TTY detection, which is why it's made a mess of the iOS logs; not sure if we can pull back the "color, but not TTY" aspect easily.

So, Rich is simply [short-circuiting](https://github.com/Textualize/rich/blob/6d30ad0f30028210124c149811cbbe2b183711f9/rich/console.py#L952) its "is tty" logic when `FORCE_COLOR` is set.

I'm somewhat hesitant to introduce our own detection since it's only really relevant to Rich. However, a call to `sys.stdout.isatty()` is pretty straightforward. The main risk would be inappropriately disabling dynamic Rich elements; although, I guess we could just default `--no-input` to the inverse of `isatty()`....not sure if that would break the piping to `stdin` we're currently using in CI...

Personally, I'd rather let Rich do what it does but allow users to turn it off in non-interactive sessions.

> If we can't - the only practical use of `--no-input` is in a CI configuration, so I don't fundamentally object to `--no-input` being overloaded.

It's also _very_ useful for quickly creating the helloworld app :)

freakboy3742:
> I'm somewhat hesitant to introduce our own detection since it's only really relevant to Rich. However, a call to `sys.stdout.isatty()` is pretty straightforward. The main risk would be inappropriately disabling dynamic Rich elements; although, I guess we could just default `--no-input` to the inverse of `isatty()`....not sure if that would break the piping to `stdin` we're currently using in CI...
> 
> Personally, I'd rather let Rich do what it does but allow users to turn it off in non-interactive sessions.

Agreed. This isn't an area where need reinvent wheels. Piggybacking on `--no-input`, or, in the extreme case, a separate `--no-animations` option seems more than enough.
 
> > If we can't - the only practical use of `--no-input` is in a CI configuration, so I don't fundamentally object to `--no-input` being overloaded.
> 
> It's also _very_ useful for quickly creating the helloworld app :)

True - but in that case, you probably also want the progress bars so you can see the downloads happening.

freakboy3742 (approved):
👍

PR #39, opened 2023-05-28
Create separate PRs for individual `pre-commit` hooks

rmartin16:
## PR Checklist:

### DIFF
--- a/.github/workflows/pre-commit-update.yml
+++ b/.github/workflows/pre-commit-update.yml
@@ -23,7 +23,7 @@ permissions:
   pull-requests: write
 
 env:
-  BRANCH_NAME: "autoupdates/pre-commit"
+  BRANCH_PREFIX: "autoupdates/pre-commit"
   CHANGENOTE_DIR: "./changes"
   FORCE_COLOR: "1"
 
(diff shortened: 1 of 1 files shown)


freakboy3742:
I presume this is motivated by docformatter 1.7.1 being a mess, but there potentially being a need for other pre-commit updates?

rmartin16:
Immediate motivation is docformatter right now. More generally, though, this approach probably makes more sense.

rmartin16:
I tested this in my [rmartin16/qbittorrent-api](https://github.com/rmartin16/qbittorrent-api/actions/runs/5106810442) repo; since `workflow_dispatch` is enabled for this workflow, once you ensure the existing `autoupdates/pre-commit` branch is deleted in the repo, you can run it manually.

rmartin16:
Although, given this system isn't as robust as dependabot, this makes it harder to test if updates to multiple hooks are compatible since this won't automatically rebase when one of the PRs is merged. This could be worked around by manually running the workflow after each merge. Alternatively, if there's perceived value in going this direction, I could see if a [path filter](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) might allow for automatically running the workflow when the `.pre-commit-config.yaml` file is touched.

freakboy3742:
> Although, given this system isn't as robust as dependabot, this makes it harder to test if updates to multiple hooks are compatible since this won't automatically rebase when one of the PRs is merged. This could be worked around by manually running the workflow after each merge. Alternatively, if there's perceived value in going this direction, I could see if a [path filter](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) might allow for automatically running the workflow when the `.pre-commit-config.yaml` file is touched.

FWIW, I agree having chained PRs would be desirable here, but given the frequency of updates we're seeing with pre-commit triggers (and, more notably, the frequency with which changes to a pre-commit tool *don't* result in any code changes that collide), I'm not sure it's worth the effort setting it up.

Is this PR ready for review, or are you anticipating additional changes?

rmartin16:
> Is this PR ready for review, or are you anticipating additional changes?

It's ready now. (Found a small typo after I marked ready for review)

Small note: because of how the name of the hook is derived (i.e. `${REPO##*/}`), the repo URL must _not_ end in a slash.

freakboy3742 (approved):
Looks good to me. Might be nice if we could get the new version number in the change note, but it's also a misc change note, so I don't know it's worth the churn.

Neat trick with the JSON->matrix handling, BTW.

PR #40, opened 2023-06-03
Clear the `JAVA_HOME*` and `ANDROID*` environment variables prior to building Android app

rmartin16:
## Changes
- GitHub runners bring along their own JDK and Android SDK
  - `JAVA_HOME=/usr/lib/jvm/temurin-11-jdk-amd64`
  - `JAVA_HOME_11_X64=/usr/lib/jvm/temurin-11-jdk-amd64`
  - `JAVA_HOME_8_X64=/usr/lib/jvm/temurin-8-jdk-amd64`
  - `JAVA_HOME_17_X64=/usr/lib/jvm/temurin-17-jdk-amd64`
  - `ANDROID_SDK_ROOT=/usr/local/lib/android/sdk`
  - `ANDROID_HOME=/usr/local/lib/android/sdk`
  - `ANDROID_NDK=/usr/local/lib/android/sdk/ndk/25.2.9519653`
  - `ANDROID_NDK_LATEST_HOME=/usr/local/lib/android/sdk/ndk/25.2.9519653`
  - `ANDROID_NDK_HOME=/usr/local/lib/android/sdk/ndk/25.2.9519653`
- NOTE: Not all GitHub runners use the same Android SDK and JDK; for instance, Ubuntu-22.04 is exposing JDK 11 while Windows Server 2022 and macOS 12.6.5 are exposing JDK 8.
- This PR clears all environment variables that start with `JAVA_HOME` and `ANDROID` so Briefcase will install and use its own Android SDK and JDK for testing.
  - The fact that all these environment variables are affecting the behavior of Briefcase and its underlying tools so much suggests that we may be better off running these tests in isolated environments like `tox` does
- Tested these changes with JDK 17: https://github.com/beeware/briefcase-android-gradle-template/actions/runs/5165357682

## Notes
- ~~A question hit me in the shower: "what will happen to people with `JAVA_HOME` already set to JDK 11?"~~
- ~~Well....turns out we've already been testing that all along since our CI has been using GitHub's JDK 11 install.~~

## PR Checklist:

### DIFF
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -277,8 +277,14 @@ jobs:
         && contains(fromJSON('["", "Gradle"]'), inputs.target-format)
         && startsWith(inputs.framework, 'toga')
       run: |
+        # force Briefcase to use its own installs of JDK and Android SDK
+        # clear all env vars starting with "JAVA_HOME" or "ANDROID"
+        unset "${!JAVA_HOME@}"
+        unset "${!ANDROID@}"
+
         cd tests/apps/verify-${{ inputs.framework }}
-        briefcase create android gradle
+        # pipe "yes" to accept Android SDK licenses
+        (2>/dev/null yes || true) | briefcase create android gradle
         briefcase build android gradle
         briefcase package android gradle --adhoc-sign
 


mhsmith:
> * A question hit me in the shower: "what will happen to people with `JAVA_HOME` already set to JDK 11?"
> * Well....turns out we've already been testing that all along since our CI has been using GitHub's JDK 11 install.

Why didn't it hit [this version check](
https://github.com/beeware/briefcase/blob/2e0fe0941f50ac0554b360af59554e2ff93dd036/src/briefcase/integrations/java.py#L105-L116)? I'd expect to see that message in [the recent logs](https://github.com/beeware/briefcase/actions/runs/5116317686/jobs/9198455529#step:24:1), but it isn't there.

rmartin16:
> > * A question hit me in the shower: "what will happen to people with `JAVA_HOME` already set to JDK 11?"
> > * Well....turns out we've already been testing that all along since our CI has been using GitHub's JDK 11 install.
> 
> Why didn't it hit [this version check](https://github.com/beeware/briefcase/blob/2e0fe0941f50ac0554b360af59554e2ff93dd036/src/briefcase/integrations/java.py#L105-L116)? I'd expect to see that message in [the recent logs](https://github.com/beeware/briefcase/actions/runs/5116317686/jobs/9198455529#step:24:1), but it isn't there.

Ahh...that's a good point. I was under the impression that it was already possible to manually specify and use a JDK 11 install with Briefcase but that check will certainly prevent that.

Therefore, we don't need to worry about anyone using JDK 11 presently...because they can't.

As for the missing error message, it's because of caching. Once Briefcase detects the JDK at `JAVA_HOME` isn't JDK 8, it goes to inform the user and install its own JDK. However, if Briefcase has already installed JDK 8 in to its data directory, it will skip showing the message about ignoring `JAVA_HOME` to the user. This is what happens in CI since we cache the Briefcase data directory.

Nonetheless, Briefcase is giving precedence to the GitHub installs of JDK and Android SDK.
- For Android SDK, it is always using GitHub's install.
- For JDK, it uses GitHub's install when it is a JDK 8....which is surprisingly inconsistent across GitHub's runner OSs.

I did some evaluation and testing of all this [here](https://github.com/rmartin16/briefcase-test-1/actions/runs/5169546778) where I printed all the environment variables, poked at the JDK at `JAVA_HOME`, and tried to initialize `Java` in Briefcase with some added logging.

freakboy3742:
What's the underlying motivation for this change? My main concern is that it will significantly slow down builds - every build will need to download the Android SDK, which... isn't small. Android builds are already painfully slow; I'd rather not make them even slower unless there's an underlying reason to do so.

rmartin16:
> What's the underlying motivation for this change?

Just to better replicate what actually happens in user installs. Although, I don't have evidence to suggest we should expect issues.

> My main concern is that it will significantly slow down builds - every build will need to download the Android SDK, which... isn't small. Android builds are already painfully slow; I'd rather not make them even slower unless there's an underlying reason to do so.

That's definitely a concern. Since we're caching the Briefcase data directory, I'd hope some of that could be mitigated....although, I'm not sure if the SDK downloaded everything there.

To that end, maybe it would be better to just explicitly set `JAVA_HOME` to whatever is in `JAVA_HOME_17_X64` so the JDK isn't downloaded either for any platform.

freakboy3742:
> > What's the underlying motivation for this change?
> 
> Just to better replicate what actually happens in user installs. Although, I don't have evidence to suggest we should expect issues.

Unless we're actually seeing problems, I'm inclined to ignore this change. Testing Briefcase works using Github CI configurations is a good canary (since we actively encourage the use of GitHub CI for cross-platform building); and it's a decent representation of the sort of developer configuration we're going to see in the "real world".

rmartin16:
That's fair. This PR was mostly another "here's an idea" PR; I always intend to make this more explicit in the PR description but then don't....I'll try to make this clearer going forward.

PR #50, opened 2023-09-11
Remove AppImage from the CI testing matrix.

freakboy3742:
AppImage builds are the slowest, because they're incompatible with binary wheels; and installing Linux GUI toolkits (which are on a constant "install the latest" push) is fundamentally incompatible with using an old base image. 

As of today, [it's impossible to install Toga on *any* manylinux image](https://gitlab.gnome.org/GNOME/pygobject/-/issues/590); PySide2 can't be installed on any *supported* manylinux image, and PySide6 only works on 2_28. Even when the install *does* work, there are so many incompatibility and binary dependency issues that it's just not worth the oxygen to keep AppImage alive.

## PR Checklist:

### DIFF
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -245,16 +245,28 @@ jobs:
         briefcase build linux system --target archlinux:latest
         briefcase package linux system --target archlinux:latest --adhoc-sign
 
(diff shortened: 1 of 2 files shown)


rmartin16 (approved):
🫡 so long appimage....you barely worked but we know you tried your best.

PR #58, opened 2023-10-24
Allow AppImage verification when explicitly requested

rmartin16:
## PR Checklist:


freakboy3742 (approved):
It took me a moment to see why this was "only when explicitly selected", and not just a reversal of the old patch. 

This should probably revert the CI config that tests AppImage on this repo as well... but there's a chicken-and-egg problem. CI won't pass until the AppImage template has been updated to downgrade the support package, and the support package won't run CI until this PR is merged. I'll add CI in a followup PR.

PR #59, opened 2023-10-24
Add support to test Flatpak with PyGame

rmartin16:
Previously, the app verification job for Flatpak would only run for Toga; this allows it to run for PyGame as well.

Alternatively, to avoid this in the future, it may be better to remove this check against the "framework" altogether and instead require the repo running the job to exclude invalid frameworks.

## PR Checklist:

### DIFF
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -273,7 +273,7 @@ jobs:
         startsWith(inputs.runner-os, 'ubuntu')
         && contains(fromJSON('["", "Linux"]'), inputs.target-platform)
         && contains(fromJSON('["", "Flatpak"]'), inputs.target-format)
-        && startsWith(inputs.framework, 'toga')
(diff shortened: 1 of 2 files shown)


freakboy3742 (approved):
All makes sense - and I agree it might be better to make the default "run everywhere", and then make the individual users of the template filter the specific frameworks that are tested.

rmartin16:
> All makes sense - and I agree it might be better to make the default "run everywhere", and then make the individual users of the template filter the specific frameworks that are tested.

Just fyi....this actually may be a bit more involved than I initially thought.

Presently, the primary consideration for whether a build is attempted for a particular output format is the OS of the GitHub CI runner. And then only Toga is allowed for the mobile platforms and PySide2 is excluded for several.

So, I think `app-build-verify.yml` may require a new input to explicitly specify the output formats to test for the a framework....instead of the inferring it from the runner OS.

PR #64, opened 2023-11-09
Support template branch input `app-build-verify` workflow

rmartin16:
## Changes
- The `app-build-verify` workflow currently supports an input for `--template` for `briefcase new`
- This adds the ability to specify `--template-branch` as well

## Notes
- Can be merged independently of the GUI framework plugin support
- Currently using for testing https://github.com/beeware/briefcase/pull/1524

## PR Checklist:

- [X] All new features have been tested
- [X] All new features have been documented
- [X] I have read the **CONTRIBUTING.md** file
- [X] I will abide by the code of conduct

PR #65, opened 2023-11-12
Fallback to GitHub Python if Windows store Python download fails

rmartin16:
## Changes
- Cloudflare started blocking access to the rg-adguard API from GitHub runners
  - See https://github.com/beeware/.github/pull/64
- This implements a temporary workaround to fallback to installing GitHub's Python to avoid breaking dependent workflows

## Notes
- Workflows in several repos depend on this action to succeed...so, this serves as at least a bandaid to avoid breaking CI in a bunch of places...especially given this will start failing downstream as soon as a cache miss happens...which is guaranteed at the end of the month.
- This obviously isn't a good long-term solution, though

## PR Checklist:

### DIFF
--- a/.github/actions/install-win-store-python/action.yml
+++ b/.github/actions/install-win-store-python/action.yml
@@ -7,6 +7,11 @@ inputs:
     default: "3.12"  # should default to the latest Python
     required: false
 
+outputs:
+  win-store-install-outcome:
+    description: "Outcome of the Windows Store Python install"
+    value: ${{ steps.install.outcome }}
+
 runs:
   using: composite
   steps:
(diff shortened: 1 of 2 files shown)


freakboy3742:
Hrm... I'm not sure I agree with this one. I'm not sure there's much point pretending we're testing the Windows Store version if we can't download it in the first place. Running "windows store" tests with Github Python is going to give the mistaken impression that we're actually testing something that we're not.

It seems like temporarily disabling those tests would be a better approach here - if only because it will result in less resources needed by test runs.

rmartin16:
Yeah; in principle, I agree with that. I only created this as a crutch to allow for establishing a long term solution without having to update the CI in 5 different repos....and then undo those updates if this is eventually resolved. That's not that much work but just wanted to present this as another option.

Speaking of long term solutions...do you have interest in maintaining this testing without this now blocked API? I imagine that would entail putting copies of the Appx files somewhere that CI can access.

freakboy3742:
Avoiding the re-work of rolling this back definitely makes sense. 

As for whether it's worth keeping the test at all? Honestly... while more testing is generally a good thing, I'm wondering if it might be overkill in this case. We're essentially running the test because of one very specific test of path management on App-store Python - but outside that test, App Store Python *should* be identical. While it's nice to have a real environment test of that, I do wonder if it's worth the headache. There's a bunch of other environment-specific stuff that we don't test (e.g., discrepancies between Xcode versions, to pick just one example), and in those cases, we're surviving with mocked tests. If Cloudflare et al are going to make our lives difficult, and it's not a temporary "whoops, we configured the CDN wrong", I'm inclined to say that mocks may be good enough for App-store Python tests, too.

PR #68, opened 2023-11-15
 Add reusable actions to create and/or verify a new project 

rmartin16:
## Changes
- Introduces a composite action to create a Briefcase project based on a GUI toolkit and optional template
- Introduces a workflow, `app-create-verify.yml`, for repos to run in CI that have impacts on project creation
  - Notably, `beeware/briefcase-template` and `beeware/briefcase` should implement this

BRIEFCASE_REPO: https://github.com/rmartin16/briefcase
BRIEFCASE_REF: gui-plugin-support
BRIEFCASE_TEMPLATE_REPO: https://github.com/rmartin16/briefcase-template
BRIEFCASE_TEMPLATE_REF: gui-plugin-support

## PR Checklist:

### DIFF
--- a/.github/actions/install-briefcase/action.yml
+++ b/.github/actions/install-briefcase/action.yml
@@ -38,8 +38,8 @@ runs:
         GITHUB_TOKEN: ${{ github.token }}
       shell: bash
       run: |
-        # Source version from Body of PR (if there is one; if there isn't,
-        # github.event.pull_request.number will be empty)
+        # Check Body of PR for Briefcase version to use
+        # (only PRs will have a value for github.event.pull_request.number)
         PR_BODY="${{ inputs.testing-pr-body }}"
         if [[ -z "${PR_BODY}" && -n "${{ github.event.pull_request.number }}" ]]; then
           PR_BODY=$(gh api repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }} --jq '.body')
@@ -58,10 +58,8 @@ runs:
         fi
 
         # Expose repo and version via outputs
-        echo "Briefcase Repo: ${BRIEFCASE_REPO}"
-        echo "Briefcase Ref: ${BRIEFCASE_REF}"
-        echo "repo=${BRIEFCASE_REPO}" >> ${GITHUB_OUTPUT}
-        echo "ref=${BRIEFCASE_REF}" >> ${GITHUB_OUTPUT}
+        echo "repo=${BRIEFCASE_REPO}" | tee -a ${GITHUB_OUTPUT}
+        echo "ref=${BRIEFCASE_REF}" | tee -a ${GITHUB_OUTPUT}
 
     - name: Derive Target Briefcase Version
       id: briefcase-derived
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -120,12 +120,18 @@ jobs:
       # if: ${{ !endsWith(inputs.repository, 'briefcase') }}
       uses: ./beeware-.github/.github/actions/install-briefcase
 
+    - name: Determine Briefcase Template Path
+      # If the briefcase-template repo is being tested, use the current checkout
+      id: template
+      if: endsWith(inputs.repository, 'briefcase-template')
+      run: echo "path=${{ github.workspace }}" | tee -a ${GITHUB_OUTPUT}
+
     - name: Create Briefcase Project
       id: create
       uses: ./beeware-.github/.github/actions/app-create
       with:
         framework: ${{ inputs.framework }}
-        briefcase-template-source: ${{ inputs.briefcase-template-source }}
+        briefcase-template-source: ${{ inputs.briefcase-template-source || steps.template.outputs.path }}
         briefcase-template-branch: ${{ inputs.briefcase-template-branch }}
 
     - name: Determine Output Format Configuration
@@ -134,9 +140,8 @@ jobs:
       if: endsWith(inputs.repository, '-template') && inputs.repository != 'beeware/briefcase-template'
       id: output-format
       run: |
-        # The current repo is cloned in to the current directory; that will
-        # be up levels up in the file system from the root of the app project
-        echo "template-override=--config 'template=\"../../\"'" | tee -a ${GITHUB_OUTPUT}
+        # Since the current repo is cloned in to the workspace, use it for the template
+        echo "template-override=$(printf -- "--config template=%q" "'${{ github.workspace }}'")" | tee -a ${GITHUB_OUTPUT}
 
     # In the steps below, using the builtin functions for comparison (instead of ==)
     # allows for case-insensitivity to the inputs for the workflow.
@@ -148,7 +153,8 @@ jobs:
         && contains(fromJSON('["", "app"]'), inputs.target-format)
       working-directory: ${{ steps.create.outputs.project-path }}
       run: |
-        briefcase create macOS app ${{ steps.output-format.outputs.template-override }}
+        briefcase create macOS app \
+          ${{ steps.output-format.outputs.template-override }}
         briefcase build macOS app
         briefcase package macOS app --adhoc-sign
 
@@ -159,7 +165,8 @@ jobs:
         && contains(fromJSON('["", "Xcode"]'), inputs.target-format)
       working-directory: ${{ steps.create.outputs.project-path }}
       run: |
-        briefcase create macOS Xcode ${{ steps.output-format.outputs.template-override }}
+        briefcase create macOS Xcode \
+          ${{ steps.output-format.outputs.template-override }}
         briefcase build macOS Xcode
         briefcase package macOS Xcode --adhoc-sign
 
@@ -172,7 +179,8 @@ jobs:
         WIX: ""  # force Briefcase to install and use its own version of WiX
       working-directory: ${{ steps.create.outputs.project-path }}
       run: |
-        briefcase create windows app ${{ steps.output-format.outputs.template-override }}
+        briefcase create windows app \
+          ${{ steps.output-format.outputs.template-override }}
         briefcase build windows app
         briefcase package windows app --adhoc-sign
 
@@ -185,7 +193,8 @@ jobs:
         WIX: ""  # force Briefcase to install and use its own version of WiX
       working-directory: ${{ steps.create.outputs.project-path }}
       run: |
-        briefcase create windows VisualStudio ${{ steps.output-format.outputs.template-override }}
+        briefcase create windows VisualStudio \
+          ${{ steps.output-format.outputs.template-override }}
         briefcase build windows VisualStudio
         briefcase package windows VisualStudio --adhoc-sign
 
@@ -200,7 +209,8 @@ jobs:
         sudo apt update -y
         sudo apt install -y --no-install-recommends python3-dev python3-pip libcairo2-dev libgirepository1.0-dev
 
-        briefcase create linux system ${{ steps.output-format.outputs.template-override }}
+        briefcase create linux system \
+          ${{ steps.output-format.outputs.template-override }}
         briefcase build linux system
         briefcase package linux system --adhoc-sign
 
@@ -211,7 +221,8 @@ jobs:
         && contains(fromJSON('["", "system"]'), inputs.target-format)
       working-directory: ${{ steps.create.outputs.project-path }}
       run: |
-        briefcase create linux system --target debian:bullseye ${{ steps.output-format.outputs.template-override }}
+        briefcase create linux system --target debian:bullseye \
+          ${{ steps.output-format.outputs.template-override }}
         briefcase build linux system --target debian:bullseye
         briefcase package linux system --target debian:bullseye --adhoc-sign
 
@@ -222,7 +233,8 @@ jobs:
         && contains(fromJSON('["", "system"]'), inputs.target-format)
       working-directory: ${{ steps.create.outputs.project-path }}
       run: |
-        briefcase create linux system --target fedora:37 ${{ steps.output-format.outputs.template-override }}
+        briefcase create linux system --target fedora:37 \
+          ${{ steps.output-format.outputs.template-override }}
         briefcase build linux system --target fedora:37
         briefcase package linux system --target fedora:37 --adhoc-sign
 
@@ -233,7 +245,8 @@ jobs:
         && contains(fromJSON('["", "system"]'), inputs.target-format)
       working-directory: ${{ steps.create.outputs.project-path }}
       run: |
-        briefcase create linux system --target archlinux:latest ${{ steps.output-format.outputs.template-override }}
+        briefcase create linux system --target archlinux:latest \
+          ${{ steps.output-format.outputs.template-override }}
         briefcase build linux system --target archlinux:latest
         briefcase package linux system --target archlinux:latest --adhoc-sign
 
@@ -261,12 +274,12 @@ jobs:
         # so, the version will need to be constrained to successfully build an AppImage with Toga.
         # Furthermore, Toga>0.3.1 requires PyGObject>=3.46.0 so its version is constrained as well.
         if [ "${{ startsWith(inputs.framework, 'toga') }}" = "true" ]; then
-          CONFIG_OVERRIDE_REQUIRES='requires=["toga-gtk==0.3.1", "PyGobject<3.46.0"]'
+          CONFIG_OVERRIDE_REQUIRES='--config requires=["toga-gtk==0.3.1","PyGobject<3.46.0"]'
         fi
 
         briefcase create linux AppImage \
           ${{ steps.output-format.outputs.template-override }} \
-          --config $(printf '%s' $CONFIG_OVERRIDE_REQUIRES)
+          ${CONFIG_OVERRIDE_REQUIRES}
         briefcase build linux AppImage
         briefcase package linux AppImage --adhoc-sign
 
@@ -281,7 +294,8 @@ jobs:
         sudo apt update -y
         sudo apt install -y --no-install-recommends flatpak flatpak-builder elfutils
 
-        briefcase create linux flatpak ${{ steps.output-format.outputs.template-override }}
+        briefcase create linux flatpak \
+          ${{ steps.output-format.outputs.template-override }}
         briefcase build linux flatpak
         briefcase package linux flatpak --adhoc-sign
 
@@ -292,7 +306,8 @@ jobs:
         && startsWith(inputs.framework, 'toga')
       working-directory: ${{ steps.create.outputs.project-path }}
       run: |
-        briefcase create android gradle ${{ steps.output-format.outputs.template-override }}
+        briefcase create android gradle \
+          ${{ steps.output-format.outputs.template-override }}
         briefcase build android gradle
         briefcase package android gradle --adhoc-sign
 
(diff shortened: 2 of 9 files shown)


rmartin16:
This really creates some chicken and egg problems....

rmartin16:
So, I'm trying to get to a place where none of the downstream repos need to have project creation config or pre-rolled out apps for CI. I've removed the need for the config...but the rolled apps will be more difficult since they include specific `pyproject.toml` changes to build the app using the current ref of the repo you're trying to test.

This really seems quite ripe for a plugin that adds `template = ../../` to each app format table....

freakboy3742:
> So, I'm trying to get to a place where none of the downstream repos need to have project creation config or pre-rolled out apps for CI. I've removed the need for the config...but the rolled apps will be more difficult since they include specific `pyproject.toml` changes to build the app using the current ref of the repo you're trying to test.
> 
> This really seems quite ripe for a plugin that adds `template = ../../` to each app format table....

I was literally about to break ground on this feature today. tl;dr `briefcase create -C template=...` so you can override individual app settings at runtime.

rmartin16:
Status
- This is ready for at least an initial review
- We'll need https://github.com/beeware/briefcase/pull/1542 to ensure `briefcase create` is being called correctly....but the AppImage test in `app-build-verify` demonstrates what this'll look like.

.github/workflows/app-create-verify.yml
```diff
@@ -0,0 +1,129 @@
...
+        assert pyproject_toml.is_file()
+        pprint(toml.load(pyproject_toml))
+
+    - name: Install Dependencies (Ubuntu)
+      if: startsWith(inputs.runner-os, 'ubuntu')
+      run: sudo apt update -y && sudo apt install -y --no-install-recommends libgirepository1.0-dev
+
+    - name: Test Project Build
+      working-directory: ${{ steps.create.outputs.project-path }}
+      run: briefcase build
```
rmartin16, line 122: This was mentioned previously so I included it....but it introduces a bunch of complications and both `beeware/briefcase-template` and `beeware/briefcase` run `app-build-verify.yml` which will perform app builds to a much more thorough extent. So, I'm leaning towards taking this out.

freakboy3742 (changes requested):
This mostly looks good; a couple of questions and clarifications inline.

.github/workflows/app-build-verify.yml
```diff
@@ -254,28 +257,30 @@ jobs:
...
       if: >
         startsWith(inputs.runner-os, 'ubuntu')
         && contains(fromJSON('["Linux"]'), inputs.target-platform)
         && contains(fromJSON('["AppImage"]'), inputs.target-format)
+      working-directory: ${{ steps.create.outputs.project-path }}
       run: |
-        cd tests/apps/verify-${{ inputs.framework }}
-        briefcase create linux AppImage
+        briefcase create linux AppImage \
+          ${{ steps.output-format.outputs.template-override }} \
```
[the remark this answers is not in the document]
rmartin16: They all need it; I will add them once https://github.com/beeware/briefcase/pull/1542 is merged. If I specify `--config` now, everything will fail so I only did it for AppImage since it always fails without `--config`.

.github/workflows/app-create-verify.yml
rmartin16: > the issue is that the test is running on Py3.12, which is completely fine for generating the template, but not for _building_ on Linux because of the system python dependency?

Right; running `briefcase build` basically imposes all of the restrictions of building that are already accommodated in `app-build-verify` and both these repos already run.

.github/workflows/app-build-verify.yml
```diff
@@ -254,28 +257,30 @@ jobs:
...
+          --config 'requires=["toga-gtk==0.3.1", "PyGobject<3.46.0"]'
```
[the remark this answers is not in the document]
rmartin16: Yeah; good point. From going through all the templates today, AppImage is only one I saw with a special additional configuration requirement beyond specifying `template=../../`.

I think for where to declare a build config override like this....I'm leaning towards here in `app-build-verify`....otherwise, any repo that needed to invoke a build with special requirements would need to be aware of such requirements....I'll consider it more though.

.github/actions/install-briefcase/action.yml
```diff
@@ -58,10 +58,8 @@ runs:
         fi
 
         # Expose repo and version via outputs
-        echo "Briefcase Repo: ${BRIEFCASE_REPO}"
-        echo "Briefcase Ref: ${BRIEFCASE_REF}"
-        echo "repo=${BRIEFCASE_REPO}" >> ${GITHUB_OUTPUT}
-        echo "ref=${BRIEFCASE_REF}" >> ${GITHUB_OUTPUT}
+        echo "repo=${BRIEFCASE_REPO}" | tee -a ${GITHUB_OUTPUT}
+        echo "ref=${BRIEFCASE_REF}" | tee -a ${GITHUB_OUTPUT}
```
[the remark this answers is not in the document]
rmartin16: I know right....it finally dawned on me after looking at enough repetitions of `echo` to `stdout` and then `echo` to file.

.github/workflows/app-build-verify.yml
```diff
@@ -254,29 +259,38 @@ jobs:
...
+        # PyGObject>=3.46.0 requires a version of glibc that isn't available in manylinux images;
+        # so, the version will need to be constrained to successfully build an AppImage with Toga.
+        # Furthermore, Toga>0.3.1 requires PyGObject>=3.46.0 so its version is constrained as well.
+        if [ "${{ startsWith(inputs.framework, 'toga') }}" = "true" ]; then
+          CONFIG_OVERRIDE_REQUIRES='requires=["toga-gtk==0.3.1", "PyGobject<3.46.0"]'
+        fi
+
+        briefcase create linux AppImage \
+          ${{ steps.output-format.outputs.template-override }} \
+          --config $(printf '%s' $CONFIG_OVERRIDE_REQUIRES)
```
rmartin16, line 278: I tried a lot of iterations of escape sequences to handle this more straightforwardly....but none of them were working for me and I landed here. Definitely open to improvement.

.github/workflows/app-build-verify.yml
```diff
@@ -100,144 +101,148 @@ jobs:
           ${{ steps.dirs.outputs.pip-cache-dir }}
           ${{ steps.dirs.outputs.docker-cache-dir }}
```
rmartin16, line 102: We can probably be more thoughtful in this caching.

For one, `beeware/.github` is [hitting](https://github.com/beeware/.github/actions/caches?query=sort%3Asize-desc) the 10GB cache storage limit....and Briefcase has several of these [caches](https://github.com/beeware/briefcase/actions/caches?query=sort%3Asize-desc) that are all nearly the same size....implying they are likely duplicates.

## tl;dr
I've provided rationale to only cache the Briefcase directory here (and `pip` using `actions/setup-python`).

Additionally, the cache key will be updated to remove inclusion of the GUI toolkit and add the current month and year.

## Caching `pip`
We don't need to manage the `pip` cache; `actions/setup-python` can handle that for us; for instance:
```yaml
  - name: Setup Python
    uses: actions/setup-python@v4.7.1
    with:
      python-version: ${{ inputs.python-version }}
      cache: pip
      cache-dependency-path: |
        **/setup.cfg
        **/pyproject.toml
```

## Caching `~/cookiecutters`
We're caching the downloaded cookiecutters. On its face, this sounds like a benefit....but the cache doesn't just materialize in the runner...it has to be downloaded. So, what's the difference in downloading the cache from GitHub or using `git` to download the cookiecutter on-demand from GitHub. These are effectively equivalent actions but one doesn't count against the 10GB cache limit. Additionally, only the macOS app template repo really has any heft to it anyway...so the benefit here is pretty marginal regardless.

## Stale caches
Third, when a workflow job has a cache hit, it doesn't upload any changes committed to that cache after the run. Looking at Briefcase's caches, some of them are 9 months old. That means we're downloading a cache created 9 months ago and using that as a base to run CI.

This is probably negating at least some of the desired benefits of caching here since a bunch of old data is (a) going to need to be updated during CI anyway or (b) degrades the fidelity of the testing process because we're using a Briefcase directory that was created a long time ago. For instance, right now, Briefcase CI has to re-download and install Android Commandline Tools 9.0 every run because it isn't in the current version of the cache.

To mitigate this, I can add the month and year to the cache key so they are evicted regularly. Unfortunately, this will create effectively duplicate entries among the caches over time. In turn, that means the repo will eventually always be up against its 10GB limit and GitHub will evict the least recently used entry. In general, this is bad because entries that are useful but not used often may be evicted unnecessarily. However, I don't think we really use caching like that anywhere; basically all of the caches are used in every CI run....so, this probably isn't too big of a deal.

An alternative approach might be to fingerprint the Briefcase directory; perhaps a recursive hash of all the files. This hash would then become a part of the cache key ensuring a new cache is created anytime something in the directory changes. This seems a bit over-engineered (and probably slow)....and I haven't completely thought through if files might change in the Briefcase directory as part of normal operation.

## Docker caching
We're including Docker caches except for on Linux (because GitHub doesn't support it on Linux)...but Docker is only used on Linux in CI. That said, even if we could cache Docker, the manylinux images start at over 1GB....so, it probably wouldn't even be reasonable to put them in GH cache given the 10GB limit.

That said, GitHub provides an unlimited storage amount of images published to `ghcr.io`. So, if we did want to entertain Docker caching, we could simply push the final image to GHCR and load it prior to running CI.

I'm not going to attempt that in this PR, though.

## Defining the cache key
This is the current key:
```
briefcase-${{ runner.os }}-${{ inputs.repository }}-${{ inputs.framework }}-${{ inputs.target-platform }}-${{ inputs.target-format }}
```
It basically ensures a new cache is created for any permutation of invoking this workflow. However, which of these inputs really contribute to a actually unique cache?

- `runner.os`: keep
  - Caches will certainly vary by OS
- `inputs.repository`: keep
  - Cache will be different for Android on Linux than AppImage
  - This is where updating an existing cache could be particularly useful....I'm imagining that the caches from these different runs on the same OS are all merged together to create a "super-cache"....but that starts to become really particular to this use-case
- `inputs.framework`: remove
  - Given this is really just about the Briefcase directory, it is independent of the GUI toolkit being used
- `inputs.target-platform`: keep
  - Targeting Android vs Linux is going to create different caches
- `inputs.target-format`: keep
  - Since the cache is already limited to a target platform, this is really whether the Briefcase directory is different for different formats for the same platform.
  - The Briefcase directory is generally consistent per target platform...except on Linux. And there's nothing really preventing the directory not changing further in the future for each target platform.

rmartin16: > - `inputs.framework`: remove
>     * Given this is really just about the Briefcase directory, it is independent of the GUI toolkit being used

Regarding the cache key....while the GUI toolkit does not explicitly change the contents of the Briefcase data directory, it does so implicitly through the implementation of the `app-build-verify` workflow itself.

For instance, when Briefcase's CI runs `app-build-verify` for Toga on Linux and another run for PySide6 on Linux, both of these will generate the same cache key. However, the Toga run will build an app for Android while the PySide6 run will not. Therefore, the data directories will be different but only one of them will be allowed to push the cache up for later runs.

Therefore, the GUI toolkit should remain a part of the cache key after all.

rmartin16: > For instance, right now, Briefcase CI has to re-download and install Android Commandline Tools 9.0 every run because it isn't in the current version of the cache.

While it is true that Briefcase has to download Commandline Tools 9.0 for each workflow run, it is not because the cache is stale. It is because CI is using the Android SDK that already exists on the runner; the pre-installed SDK has a version of Commandline Tools installed as `latest` so Briefcase is using that version to then install version 9.0.

This could be changed by forcing Briefcase to ignore this Android SDK and set up its own SDK that would also even be cached. However, I would expect this to become untenable given the potential size of the SDK. This will become even more so the case once I introduce `briefcase run` in to `app-build-verify` where the Android system image will be downloaded.

.github/workflows/ci.yml
```diff
@@ -19,13 +19,13 @@ env:
...
-    name: Pre-commit checks
+    name: Pre-commit Checks
     uses: ./.github/workflows/pre-commit-run.yml
     with:
-      pre-commit-source: "pre-commit"
+      pre-commit-source: pre-commit
 
   test-docs-build-verify:
-    name: Test Verify Docs Build
+    name: Verify Docs Build Check
```
freakboy3742, line 28: Minor Nit: The task is verifying the docs build; verifying a check is... a bit meta.
```suggestion
    name: Verify Docs Build
```

.github/workflows/app-create-verify.yml
```diff
@@ -0,0 +1,132 @@
...
+        repository: rmartin16/.github-beeware  # !!!! TODO:PR: REPLACE ME WITH beeware/.github !!!!
+        ref: create-new-project  # !!!! TODO:PR: REMOVE ME !!!!
+        path: beeware-.github
+
+    - name: Checkout beeware/briefcase-template
+      uses: actions/checkout@v4.1.1
+      with:
+        repository: rmartin16/briefcase-template  # !!!!! TODO:PR: REPLACE ME WITH beeware/briefcase-template !!!!!
+        ref: gui-plugin-support  # !!!!! TODO:PR REMOVE ME !!!!!
+        path: briefcase-template
```
rmartin16, line 62: It isn't clear why I added this....but it definitely isn't necessary now. It should be removed when the updates are made prior to merging.

rmartin16: Ohh nvm, i see now. This is needed to install the requirements for linting the project. So, it should stick around.

.github/workflows/app-build-verify.yml
freakboy3742: Agreed that pip and cookiecutter caches can be optimised in the way you describe. Also agreed that there's an opportunity for Docker image caching, but we don't need to solve that problem here.

I'm fairly certain the rest of your caching argument is sound; and AFAICT the implementation essentially comes down to the cache key - and the worst case is "the data you need isn't cached and needs to be downloaded on demand", so AFAICT we shouldn't see *failures* as a result of any caching issue, just longer test times. We can iterate on these over time if we need to.

Some caching related thoughts/ideas. Unless you're feeling especially inspired, I don't think we need to tackle these as part of this PR. Even if we turn *off* caching, all we see is slightly slower test runs; I'm only flagging these since the topic has come up.

1. Given the inconsistency of interpretation in `input.target-format`, I wonder if a partial solution is to make the cache key itself platform dependent. That would allow macOS to use the same key for all formats, but allow AppImage and Flatpak to have different keys.

2. Another possible approach would be to fracture the cache into smaller parts, and pull it on demand. Rather than a single "Briefcase" cache, cache the support packages, android toolchain, java toolchain, etc, and request the specific cache keys that are needed. Harder to manage, but less likely to end up with competing cache keys.

3. We've been deliberately avoiding the pre-installed tools on the CI machines... but I wonder if it might be worth relaxing that (or, at least, relaxing them as an option). The two tools that actually benefit from caching are Android SDK and Java - the remaining tools are sufficiently small that a cache miss isn't *that* big a deal. The versions provided on-machine are knowable and reliable; Java 17 is available on all images; and the Ubuntu-latest  image [ships with command line tools version 9](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md) - so we're not actually gaining anything by explicitly downloading those tools, other than exercising the download mechanism. If the default Android SDK version were to change, that would be an interesting quandary - but it would also be a useful test that we're tracking the behavior of the Android toolchain. 

If we take the Android and Java components out of the toolchain, we also remove a *lot* of our machine needs - the next biggest chunk would be the support packages, and those are pretty safe to cache on a per-OS basis.

rmartin16: > so AFAICT we shouldn't see failures as a result of any caching issue, just longer test times.

My biggest concern is undetected/unreported issues because CI is running with a stale/old cache....although, these seem probably rather unlikely. Nonetheless, I'm still planning to implement a quick job that will periodically delete all the caches so they can be re-built.

> Unless you're feeling especially inspired, I don't think we need to tackle these as part of this PR.

Beyond what I've already done, I'd rather save further cache tweaking for the future. These are definitely some good considerations.

.github/workflows/app-build-verify.yml
```diff
@@ -63,181 +63,191 @@ jobs:
...
       if: >
         startsWith(inputs.runner-os, 'ubuntu')
-        && !contains(fromJSON('["PySide2"]'), inputs.framework)
         && contains(fromJSON('["", "Linux"]'), inputs.target-platform)
         && contains(fromJSON('["", "system"]'), inputs.target-format)
+      working-directory: ${{ steps.create.outputs.project-path }}
       run: |
-        cd tests/apps/verify-${{ inputs.framework }}
-        briefcase create linux system --target fedora:37
+        briefcase create linux system --target fedora:37 \
```
freakboy3742, line 237: Doesn't need to be fixed in this PR, but flagging that Fedora 37 is EOL in 2 weeks, so we're likely going to need to bump this to 38 or 39 soon.

.github/workflows/app-build-verify.yml
```diff
@@ -254,29 +264,39 @@ jobs:
...
         startsWith(inputs.runner-os, 'ubuntu')
         && contains(fromJSON('["", "Linux"]'), inputs.target-platform)
         && contains(fromJSON('["", "Flatpak"]'), inputs.target-format)
         && contains(fromJSON('["Toga", "PyGame"]'), inputs.framework)
+      working-directory: ${{ steps.create.outputs.project-path }}
       run: |
-        sudo apt-get update -y
-        sudo apt-get install -y flatpak flatpak-builder
+        sudo apt update -y
+        sudo apt install -y --no-install-recommends flatpak flatpak-builder elfutils
```
freakboy3742, line 296: What is elfutils for? Why the new requirement?

rmartin16: I added `--no-install-recommends` to avoid all the stuff we don't need that `apt` installs by default.

As for `elfutils`, it's only a recommended package for `flatpak-builder` but it's required because it contains `eu-strip` which `flatpak-builder` needs.

freakboy3742 (approved):
Marking as approved, conditional on changes that are needed as part of the [beeware/briefcase#1524 landing strategy](https://github.com/beeware/briefcase/pull/1524#pullrequestreview-1736118477)

.github/workflows/app-build-verify.yml
rmartin16: 👍🏼 I should take care of this in my next round of PRs. I think some of the versions in the templates (e.g. for `org.kde.Platform`) need bumps as well.

freakboy3742 (approved):
Approved for final ship.

.github/workflows/app-build-verify.yml
rmartin16: > 3\. We've been deliberately avoiding the pre-installed tools on the CI machines... but I wonder if it might be worth relaxing that (or, at least, relaxing them as an option). The two tools that actually benefit from caching are Android SDK and Java - the remaining tools are sufficiently small that a cache miss isn't _that_ big a deal. The versions provided on-machine are knowable and reliable; Java 17 is available on all images; and the Ubuntu-latest  image [ships with command line tools version 9](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md) - so we're not actually gaining anything by explicitly downloading those tools, other than exercising the download mechanism. If the default Android SDK version were to change, that would be an interesting quandary - but it would also be a useful test that we're tracking the behavior of the Android toolchain.

Using the Java 17 that GitHub that provides will be trivial; it's location is in `$JAVA_HOME_17_X64`.

We actually already use the GitHub Android SDK. And it does contain Command-Line Tools 9.0 already....however, the tools are installed as `latest` and not `9.0`. I could update Briefcase to use `latest` if it is indeed 9.0....but in reality this download doesn't take long at all. Even downloading the system image to run the emulator takes <30s.

At any rate, using the GitHub Java 17 prevents Briefcase from using the data directory at all.

PR #69, opened 2023-11-18
Run apps created in `app-build-verify`

rmartin16:
## Changes
- Run `briefcase run` for apps created with `app-build-verify`
- Introduce new "Briefcase Automation" plugin
  - This plugin replaces the `app.py` with a version that will automatically exit shortly after the app starts
  - https://github.com/beeware/briefcase/pull/1549
- Use the GitHub provided WiX install
- Use the GitHub provided Java 17 install
- Use Fedora 39 to test RPM packaging (instead of Fedora 37)
- Closes https://github.com/beeware/briefcase-template/issues/75

## Notes
- With some cheats, all frameworks are running on all platforms (where frameworks are supported at all)
- `briefcase run` is not called for apps created for Linux via containerization since Briefcase does not support the Run command when `--target` is specified (and that's because passing the X11 into Docker is convoluted).

## PR Checklist:


rmartin16:
An interesting conundrum.

If another repo, like `beeware/briefcase`, invokes a version of this workflow that isn't at `beeware/.github@main`, how do you ensure that subsequent workflows and actions are also from that same non-`main` branch?

Notably, consider Briefcase's CI calling `app-build-verify` but from a branch on my fork, i.e. `rmartin16/.github-beeware/.github/workflows/app-build-verify@testing-branch`. This, of course, works fine but `app-build-verify` needs to run the `app-create` action to roll out the Briefcase project for testing. It does this by referencing the `app-create` action via a local filesystem path to the `.github` repo.....and hence, where the problem begins: how do you checkout the `.github` repo for the version of `app-build-verify` that's currently running?

This will require knowing the repo and branch where the running `app-build-verify` lives.

rmartin16:
Another person [pointed out](https://github.com/orgs/community/discussions/31054) this same issue. They did not really come to a reasonable conclusion, though...since their workable solution was a hacky javascript action that manually inspects the environment to derive where the running reusable workflow lives....and using some OIDC context at that.

The core problem is mostly how inheritance works. When the reusable workflow (i.e. `app-build-verify`) is called, it inherits the environment from the caller; this includes the `github` context.

Relevant bits of `github` while `app-build-verify` is running when Briefcase calls it in a PR:
```
github.ref_name=1546/merge
github.repository=beeware/briefcase
github.repository_owner=beeware
github.workflow=CI
github.workflow_ref=beeware/briefcase/.github/workflows/ci.yml@refs/pull/1546/merge
github.workflow_sha=37da207f9e010c326f1c0a93730e1fd02209e57c
```
So, the reusable workflow being called effectively appears to just be a part of the workflow that called it.

This means a naked `actions/checkout` step in `app-build-verify` will checkout `beeware/briefcase` and not the repo for the version of `app-build-verify` that's running.

There doesn't appear to be a straight-forward way for a reusable workflow on a non-`main` branch to naturally use other reusable workflows/actions from that same branch. Hell, when they first added reusable workflows, they apparently didn't even have way to specify workflows in the same repo without explicitly specifying the same org/repo....so, this problem initially existed even when trying to test reusable workflow changes in the same repo. Referencing relative workflows was eventually [added](https://github.blog/changelog/2022-01-25-github-actions-reusable-workflows-can-be-referenced-locally/).

Anyway, I suppose I'll add inputs to `app-build-verify` to explicitly declare the repo to use for nested workflows/actions.

rmartin16:
A few things to decide:
- Where should the Briefcase Automation plugin live?
  - Currently, its in my [github](https://github.com/rmartin16/briefcase-plugin-test)....but that obviously isn't appropriate long term
  - We can simply create a new repo under the `beeware` org
  - I'm also quite tempted to say it goes in `beeware/briefcase` as an additonal package...
- Testing strategy
  - I've gone "all the way" with this PR currently....in that it supports running (nearly) all GUI toolkits on all platforms
    - Do we think this is necessary? or desired?
  - To get PPB to run on any platform, I had to use an [old pysdl2-dll](https://github.com/rmartin16/briefcase-plugin-test/blob/main/src/briefcase_plugin_test/pursuedpybear.py) version
    - It seems like PPB is looking for symbols that just don't exist in the shared libraries in more recent packages
    - I didn't look in to it too deeply...but this may be something that should be added to the real templates
  - Does any of the environment configuration to run the apps negate meaningful testing?
    - For instance, does setting `QT_QPA_PLATFORM=offscreen` mean we're not meaningfully testing PySide6?
  - CI for `.github` is ensuring all GUI toolkits pass `app-build-verify`
    - We can go back to testing less per output format is this is just too many
- CI Impacts
  - Adding this to CI doesn't appear to add much to its running time
  - For instance, Briefcase took [1hr48m](https://github.com/beeware/briefcase/actions/runs/6910715080/usage) of CPU time prior and (an average of) [1hr56m](https://github.com/beeware/briefcase/actions/runs/6922095306/usage) after

freakboy3742:
> A few things to decide:
> 
> * Where should the Briefcase Automation plugin live?
>   * I'm also quite tempted to say it goes in `beeware/briefcase` as an additonal package...

The beeware/briefcase repo makes sense to me. It doesn't need to be installed by default, as it won't be very useful for anyone other than internal testing, but I don't think it's worth the effort of breaking it out as a standalone repo.

> * Testing strategy
>   
>   * I've gone "all the way" with this PR currently....in that it supports running all GUI toolkits on (nearly) all platforms
>     
>     * Do we think this is necessary? or desired?

Yes? Unless it's prohibitively difficult or time-expensive, I can't see any reason we shouldn't verify "this app can start" for as many targets as is plausible. We've *built* them - we might as well make sure they work.

>   * To get PPB to run on any platform, I had to use an [old pysdl2-dll](https://github.com/rmartin16/briefcase-plugin-test/blob/main/src/briefcase_plugin_test/pursuedpybear.py) version
>     
>     * It seems like PPB is looking for symbols that just don't exist in the shared libraries in more recent packages
>     * I didn't look in to it too deeply...but this may be something that should be added to the real templates

I've noticed that too in my local testing. There might be something to report upstream to @pathunstrom (or she may already be aware), or it could be a template tweak we need to apply.

>   * Does any of the environment configuration to run the apps negate meaningful testing?

I don't know - but "it starts and shuts down" is honestly a *lot* more testing than the current state of affairs. It's easy to get a binary that compiles but doesn't start. A binary that starts but doesn't do the right thing is harder, but those problems are more in the "are you using the toolkit right" territory than the "are you compiling it right" territory - and Briefcase's concern is more about the latter than the former.

rmartin16:
> > * Testing strategy
> >   
> >   * I've gone "all the way" with this PR currently....in that it supports running all GUI toolkits on (nearly) all platforms
> >     
> >     * Do we think this is necessary? or desired?
> 
> Yes? Unless it's prohibitively difficult or time-expensive, I can't see any reason we shouldn't verify "this app can start" for as many targets as is plausible. We've _built_ them - we might as well make sure they work.

There's a few output formats that are just a little iffy....so FYI if nothing else:
- AppImage (because of course AppImage)
  - I'm cheating with several of the GUI toolkits since I'm not actually ensuring the AppImage would run outside of CI....Ive just installed the necessary libraries in to the runner and the AppImage loads those from `/usr/lib`
  - This barely matters, though, since this only runs for `briefcase-linux-appimage-template` at this point
- Flatpak
  - I'm injecting `--device=dri` in to the Flatpak manifest for some of the non-Toga toolkits
- Android
  - I couldn't figure out how to get the emulator to run on Windows...but I feel like this should be possible

freakboy3742:
> * AppImage (because of course AppImage)
>   
>   * I'm cheating with several of the GUI toolkits since I'm not actually ensuring the AppImage would run outside of CI....Ive just installed the necessary libraries in to the runner and the AppImage loads those from `/usr/lib`
>   * This barely matters, though, since this only runs for `briefcase-linux-appimage-template` at this point

The extent to which I care about AppImage at this point is difficult to measure with conventional instrumentation. :-)

> * Flatpak
>   
>   * I'm injecting `--device=dri` in to the Flatpak manifest for some of the non-Toga toolkits

If it's needed, it's needed. Is there a good reason to *not* include this permission? It doesn't strike me as a bad idea to enable network, sound and graphics flags for *all* toolkits, not just non-Toga.

> * Android
> 
>   * I couldn't figure out how to get the emulator to run on Windows...but I feel like this should be possible

So - two notes here: 
1. I'm not sure it is possible - AFAIK, GitHub Actions machines are missing the kernel modification needed to run the Android emulator. This is essentially the same thing that kept us from running Android CI on Linux (until the Linux situation changed)
2. We don't need to run the Android image on every OS - we just need to verify that it *builds* on every OS, and that it runs in the emulator. As long as it runs in *an* emulator (presumably the Linux one, because of speed), I'm happy to call that a success.

freakboy3742:
> > * AppImage (because of course AppImage)
> >   
> >   * I'm cheating with several of the GUI toolkits since I'm not actually ensuring the AppImage would run outside of CI....Ive just installed the necessary libraries in to the runner and the AppImage loads those from `/usr/lib`
> >   * This barely matters, though, since this only runs for `briefcase-linux-appimage-template` at this point
> 
> The extent to which I care about AppImage at this point is difficult to measure with conventional instrumentation. :-)
> 
> > * Flatpak
> >   
> >   * I'm injecting `--device=dri` in to the Flatpak manifest for some of the non-Toga toolkits
> 
> If it's needed, it's needed. Is there a good reason to _not_ include this permission? It doesn't strike me as a bad idea to enable network, sound and graphics flags for _all_ toolkits, not just non-Toga.
> 
> > * Android
> >   
> >   * I couldn't figure out how to get the emulator to run on Windows...but I feel like this should be possible
> 
> So - two notes here:
> 
> 1. I'm not sure it is possible - AFAIK, GitHub Actions machines are missing the kernel modification needed to run the Android emulator. This is essentially the same thing that kept us from running Android CI on Linux (until the Linux situation changed)
> 2. We don't need to run the Android image on every OS - we just need to verify that it _builds_ on every OS, and that it runs in *an* emulator. As long as it runs in _an_ emulator (presumably the Linux one, because of speed), I'm happy to call that a success.

rmartin16:
> > * Flatpak
> >   
> >   * I'm injecting `--device=dri` in to the Flatpak manifest for some of the non-Toga toolkits
> 
> If it's needed, it's needed. Is there a good reason to _not_ include this permission? It doesn't strike me as a bad idea to enable network, sound and graphics flags for _all_ toolkits, not just non-Toga.

Yeah; that makes sense. Network is already enabled by default. I can add graphics (via `device=dri`) and sound (via `socket=pulseaudio`). We may also want to eventually consider allowing "running in the background" with advent of "status bar apps".

.github/workflows/app-build-verify.yml
```diff
@@ -110,24 +117,24 @@ jobs:
...
       uses: actions/download-artifact@v3.0.2
       with:
-        name: packages-${{ steps.package.outputs.name }}
+        name: packages-briefcase
         path: dist
 
     - name: Install Briefcase Artefact
       if: endsWith(inputs.repository, 'briefcase')
-      run: python -m pip install dist/briefcase-*.whl
+      run: python3 -m pip install dist/briefcase-*.whl
```
rmartin16, line 130: Realized I've been making a flawed assumptions with the gating of some of these steps.

These checks on the name of the repository just assume that forks will use the same name as the repos in the beeware org. But there's definitely zero requirement on that.

Checks like this should be updated to checking the name of the target repo.

I think the closest thing to that is the "base" of a PR accessible here:
```
${{ github.event.pull_request.base.repo.full_name }}
```
This would yield, for instance, `beeware/briefcase` for PRs for Briefcase irrespective of the name of the forked repo.

rmartin16: However, the CI workflow obviously also runs for events that aren't for PRs. So, an ultimate solution will need to fallback to other values in that situation.

(mostly notes to myself....debating about addressing this now or not....this hasn't been an actual issue yet....afaik)

freakboy3742: Agreed this is technically an issue, and I'm totally open to fixes - but I'm also not exceptionally concerned by it. The set of people who change the name of a repo on fork is... small. I and they can contribute the fix if they're bitten by this problem.

.github/actions/app-create/action.yml
```diff
@@ -65,12 +82,18 @@ runs:
...
+        case "$(tr '[:upper:]' '[:lower:]' <<< "${GUI_TOOLKIT}")" in
+          toga         ) GUI_INPUT=1 ;;
+          toga-auto    ) GUI_INPUT=8 ;;
+          pyside6      ) GUI_INPUT=2 ;;
+          pyside6-auto ) GUI_INPUT=6 ;;
+          ppb          ) GUI_INPUT=3 ;;
+          ppb-auto     ) GUI_INPUT=5 ;;
+          pygame       ) GUI_INPUT=4 ;;
+          pygame-auto  ) GUI_INPUT=7 ;;
+          *            ) GUI_INPUT=${{ inputs.framework }} ;;
```
freakboy3742, line 96: So - this *works*, but I wonder if there's a better way.

Instead of trying to program STDIN content, how about we expose the wizard options to the command line. Give each question a metadata identifier (could be as simple as the context variable that is being populated), and expose a command line option that allows the end user to specify answers to those questions - e.g., `briefcase new -Q app_name=my_app -Q formal_name="My App" -Q gui_toolkit=toga`. (I'm using `-Q/--question` here, but I'm open to other spelling)

These command-line values override the default values for each question; that means with --no-input, you'll get default values except for the values you override.

In the case of the template question, you'd be looking to provide the key from the dictionary - so the identifier of the plugin.

Not sure if this needs to be a blocker for this PR; but since it's going to need another pair of .github and briefcase paired changes, I don't know if it might be easier to include in this set.

rmartin16: Yeah...this is definitely better. I had been trying to avoid doing this because it only improves CI/dev ergonomics slightly. However, the feature isn't really that complicated.

1 - new repeatable cmdline argument that contribs to a dict
2 - check this dict before asking the question
3 - warn about any unused inputs

There is the complication of invalid inputs. Given this is just a dev thing, I'll probably keep it simple and just output a warning and keep going.

rmartin16:
note to self: interesting....Briefcase crashed out while running the app...but [didn't fail](https://github.com/beeware/briefcase/actions/runs/6985560753/job/19010015822#step:17:134) CI...although, in another example where it [did fail](https://github.com/beeware/briefcase/actions/runs/6985560753/job/19010015678#step:16:86) CI, the exception in the output streamer thread wasn't captured in the [Briefcase log](https://github.com/beeware/briefcase/suites/18482789371/artifacts/1072261004)...

.github/actions/install-briefcase/action.yml
```diff
@@ -141,6 +141,8 @@ runs:
 
         echo "Installing ${REPO}@${REF}"
         python -m pip install -U "git+${REPO}@${REF}"
+        # this command can fail for versions of Briefcase <= 0.3.16
+        python -m pip install -U "git+${REPO}@${REF}#subdirectory=automation" || true
 
         echo "Installed version: $(briefcase --version)"
         echo "version=${REF}" >> ${GITHUB_OUTPUT}
```
rmartin16, line 148: hmm....this will be an interesting problem. The `install-briefcase` action usually runs to install from `main` or from a PR branch....but it can also be used to install previous releases of Briefcase. This will happen for PRs targeting the "version branches" on the templates repos....although, these PRs are not especially common.

Nonetheless, the Automation plugin will obviously not be available for these previous versions of Briefcase. So, while this command can fail and move on, we'll need to ensure that `app-build-verify` can still successfully run. This will effectively include ensuring that `enable-automation` is `false` and that `briefcase run` does not actually run.

ideas forthcoming...

.github/actions/app-create/action.yml
rmartin16: - https://github.com/beeware/briefcase/pull/1552

.github/actions/install-briefcase/action.yml
rmartin16: This strategy works...
- Ensure `python -m pip install` succeeds whether `briefcase-automation` can be installed or not
- Before building and running the apps, confirm whether `briefcase-automation` is installed
- Only run `briefcase run` if `briefcase-automation` could be installed

.github/actions/app-create/action.yml
```diff
@@ -84,10 +89,12 @@ runs:
         rm -rf "${APP_DIR}"
 
         # Roll out the project
-        printf "%s\n%s\n\n\n\n\n\n\n\n%s\n" "${APP_NAME}" "${APP_DIR}" "${GUI_INPUT}" \
-          | briefcase new ${TEMPLATE} ${TEMPLATE_BRANCH}
+        briefcase new --no-input ${TEMPLATE} ${TEMPLATE_BRANCH} \
+          -Q "formal_name=${APP_NAME}" \
+          -Q "app_name=${APP_DIR}" \
+          -Q "bootstrap=${BOOTSTRAP}"
```
rmartin16, line 95: ahhh...of course....this won't work for previous versions of Briefcase that backport PRs for templates will be using...

this almost starts to ask the question "should we have versioned branches for CI....?"

rmartin16:
This is basically all here now....
- Except for support to call `briefcase new` for the backport PRs for the templates without using the new `-Q` argument
  - It would be pretty awkward to fall back to injecting to the inputs to `stdin`
  - I do wonder if the `app-create` action could just always use the latest Briefcase version....but that means that the proper Briefcase template version needs to be worked out in CI instead...
- I've always been quite deep in this for a while now....so fresh eyes would probably be useful regardless

.github/actions/app-create/action.yml
rmartin16: So, even if I pivot the invocation of `briefcase new` to use stdin injection for previous releases of Briefcase, there's a more fundamental problem.

When we merged #68, that was actually when we broke the ability to run CI for backport PRs for templates since we removed the expectation in `app-build-verify` for templates to contain an app to test. Instead, we started using `--config` to define the template file path to use to create an app on-the-fly. Of course, if CI is using an older release of Briefcase, `--config` is not going to work.

rmartin16: Given this, I think it would be particularly onerous at this point for CI to support previous versions of Briefcase; the complexity is already above where we'd like and this would drastically compound that.

So, I think we have two options:
1) Disable CI for template PRs that are not targeting their respective `main` branch
    - Obviously the simplest choice....and maybe the reasonable choice
    - Backport PRs are rare and are usually just version bumps in `pyproject.toml`
2) Create a new branch in the `beeware/.github` repo that exposes the old CI for template repo branches to use
    - This introduces additional complexity and tech debt....but not nearly as much as trying to support older releases of Briefcase in `main`
    - Implementation should be straightforward...although, tedious. Once a new branch off an older commit exists in `beeware/.github`, all of the versioned branches in the template repos would need to be updated to use it
    - Alternatively, we could avoid such updates until any PRs are actually created for one of those branches

rmartin16:
### Testing
- Briefcase repo
  - PR using these changes: https://github.com/beeware/briefcase/pull/1549
  - Shows `app-create-verify` _not_ [using](https://github.com/beeware/briefcase/actions/runs/7037952443/job/19154086788#step:10:122) the Automation plugin to create a Briefcase project
  - Shows `app-build-verify` [installing](https://github.com/beeware/briefcase/actions/runs/7037952443/job/19154083987#step:9:1) Briefcase from the packaged wheel
  - Shows `app-build-verify` [creating](https://github.com/beeware/briefcase/actions/runs/7037952443/job/19154083987#step:11:131) Briefcase project with Automation plugin
  - Shows `app-build-verify` [running](https://github.com/beeware/briefcase/actions/runs/7037952443/job/19154083987#step:13:109) app and exiting
- Briefcase project template repo
  - PR using these changes: https://github.com/beeware/briefcase-template/pull/86
  - Shows `app-create-verify` [using](https://github.com/beeware/briefcase-template/actions/runs/7038685523/job/19156170833#step:10:114) the local repo checkout to create the Briefcase project
  - Shows `app-build-verify` [using](https://github.com/beeware/briefcase-template/actions/runs/7038685523/job/19156167438#step:11:123) the local repo checkout to create the Briefcase project
- Briefcase build template repo
  - PR using these changes: https://github.com/beeware/briefcase-android-gradle-template/pull/74
  - Shows `app-build-verify` [using](https://github.com/beeware/briefcase-android-gradle-template/actions/runs/7038046330/job/19154221934#step:23:160) the local repo checkout to lay out the build directory

.github/actions/install-briefcase/action.yml
rmartin16: See https://github.com/beeware/.github/pull/69#discussion_r1409618409 for continued discussion of backport PRs for templates.

.github/actions/app-create/action.yml
freakboy3742: Completely agreed that excessive complexity in aid of CI on old template branches isn't worth it. 

I'll go a step further on your "backport PRs are rare" statement, and say that a backport PR for anything other than the most recent version is almost unheard of. The only times we've ever backported a patch is when we've noticed something problematic in the current stable version that can be fixed without a full Briefcase release.

That means that once 0.3.17 is in the wild, all the branches we actually care about *will* be able to run CI. It's just this intermediate stage where 0.3.16 doesn't support the extra arguments that is a problem. If that means we disable app-build-verify for non-main branches (or, better still, "non-main branches before 0.3.17"), I'd be fine with that.

With that said, it does make me wonder if the root cause of the problem is that we're encoding too much logic in this .github repo. The complexity here is mostly because the arguments being passed *to* briefcase new/create/build/run/package are embedded in *this* repo, not the invoking repo. Moving the invocation-specific logic to the calling process might involve a *little* duplication; but it should (AFAICT) significantly simplify the .github workflows, and would allow us to pass different arguments in on a CI *branch*.

rmartin16: > With that said, it does make me wonder if the root cause of the problem is that we're encoding too much logic in this .github repo. The complexity here is mostly because the arguments being passed _to_ briefcase new/create/build/run/package are embedded in _this_ repo, not the invoking repo. Moving the invocation-specific logic to the calling process might involve a _little_ duplication; but it should (AFAICT) significantly simplify the .github workflows, and would allow us to pass different arguments in on a CI _branch_.

Soo...this would effectively amount to an entire parametrization of the workflow.....such that any logic would be pushed to the calling repo. This would include:
- Source of the Briefcase project template
- Extra parameters to use to run `briefcase new`
- Source of the Briefcase build template
- Extra steps to perform (such as `apt install`) before building the app

This is all possible....probably wouldn't even be that hard (famous last words 😉)....although, I'm not sure the simplification would be particularly significant; doing this would only really remove 2 steps.....it may even require introducing another step.

There's also an opportunity to improve whether Briefcase is installed from a pre-built package or whether the version to install is derived from the nature of the CI event.

I'm willing to experiment with this idea once beeware/briefcase#1549 is merged....but I think I'm not expecting too big of changes. Please let me know, though, if you're thinking of other levels of changes.

freakboy3742: Hrm - you may be right that this won't actually simplify thing that much.

I should have a chance to do a deep review of this tomorrow morning; I'll keep this sort of thing in mind when I do that review.

rmartin16: Oh....also just remembered we'd have to parameterize whether to run `briefcase create` or use an existing project as well because `--config` won't work for older versions of Briefcase; this, of course, goes along with passing the arguments for `briefcase new`.

I guess in a lot of ways, we've fundamentally changed the expectations for this part of CI.

freakboy3742 (changes requested):
A couple of questions for clarification purposes, and one minor cosmetic thing - but otherwise, I think this all makes sense, and it's difficult to argue with the success of the example cases you've provided. 

I definitely can't see anything that is systematically wrong or will lock us out of a use case that we need to support; my only real concern is the level of complexity... but to some extent, this is just inherently complex.

.github/workflows/app-create-verify.yml
```diff
@@ -73,38 +73,33 @@ jobs:
...
-
     - name: Create Briefcase Project
       id: create
       uses: ./beeware-.github/.github/actions/app-create
       with:
         framework: ${{ inputs.framework }}
-        briefcase-template-source: ${{ inputs.briefcase-template-source || steps.template.outputs.path }}
-        briefcase-template-branch: ${{ inputs.briefcase-template-branch }}
+        # use the current repo checkout if briefcase-template is being tested
+        briefcase-template-source: ${{ endsWith(github.repository, 'briefcase-template') && github.workspace || '' }}
```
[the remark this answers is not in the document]
rmartin16: This is definitely a little more obscure; so, https://github.com/beeware/.github/pull/68 added the ability to specify the briefcase template as part of the body of the PR; so, manually overriding it in the workflow isn't necessary anymore.

rmartin16:
This is all here now.

I'm not sure where we stand now with any further parameterization of `app-build-verify`; do you think it'll be worth abstracting these [last few bits](https://github.com/beeware/.github/pull/69#discussion_r1410128415)?

Otherwise, the only other question is CI for backports for templates. I think we'll just need to abandon running CI for such PRs; I definitely do not see any low-effort option to keep it working. Although, once Briefcase 0.3.17 is cut, CI would be able to run for backports to that version and later.

freakboy3742 (approved):
All looks good to me.

Regarding the potential parameterisation: I think I'm happy for now. It isn't immediately obvious to me that parameterisation will result in something that is *obviously* better - I think we'd likely end up exchanging some complexity in this project for equal levels of complexity spread across all the templates (... or, straight up more complexity in both 😝). 

We can always revisit this if one of us has a bright idea about how to simplify everything, but for now, the perfect is the enemy of the good, and what we've got definitely meets the bar of good (and then some).

rmartin16:
I just realized that CI for this PR should be re-run without `BRIEFCASE_REPO` in the PR body. Let me do that.

PR #70, opened 2023-11-19
Add workflow to delete expired caches

rmartin16:
## Changes
- Deletes any cache that is older than 30 days once per day
- Demonstration: https://github.com/rmartin16/briefcase-arch-test/actions/runs/6922927318/job/18830228205#step:2:445

## PR Checklist:

### DIFF
--- a/.github/workflows/caches-expired-delete.yml
+++ b/.github/workflows/caches-expired-delete.yml
@@ -1,0 +1,59 @@
+name: Delete Expired Caches
+
+#######
+# Deletes caches that were created more than 30 days ago.
+#######
+
+on:
+  schedule:
+    - cron: "0 20 * * *"  # each day @ 2000 UTC
+  workflow_dispatch:
+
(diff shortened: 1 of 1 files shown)


rmartin16:
As an alternative (since I'll be the first to admit this is particularly hacky), GitHub already has support for cache pruning; any cache that hasn't been accessed in over 7 days is automatically deleted.

rmartin16:
I realize now there's a critical flaw in this hashing idea...you can't use a hash of the thing you're trying to cache as the key to retrieve the cache.....because how could you calculate the hash initially to know which cache to fetch. This approach could only be used to know whether to replace a cache after you've fetched and used it.

Ok; so, we are forced to use a proxy for describing the briefcase data directory in the key....like we're already doing.

That said, I'll construct a cache key that'll be more accurate for the target platform and also incorporates the date so they are replaced on a regular basis.

PR #71, opened 2023-11-19
Use a platform-specific cache key that rotates each month

rmartin16:
## Changes
- Only incorporates the target output format in the the cache key for Linux
- The month and year are also included so the cache is replaced each month

## Notes
- Need to ensure the data directory doesn't change

## PR Checklist:


freakboy3742 (approved):
👍

PR #79, opened 2024-01-08
Run `setup-python` if the requested Python is not the System Python

rmartin16:
## Changes
- Linux System builds require the app is built with the System Python; that is, the Python from `setup-python` cannot be used if you also want to run the created app
- Since the Linux System build is only actually run if the requested Python is the same as the System Python, run `setup-python` if they are different

Demo for Briefcase: https://github.com/beeware/briefcase/actions/runs/7454601543/job/20282336558
Demo for AppImage: https://github.com/beeware/briefcase-linux-appimage-template/actions/runs/7454506398/job/20281949158#step:7:20

## PR Checklist:

### DIFF
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -107,7 +107,7 @@ jobs:
 
     - name: Set Up Python
       # Linux System apps requires python is System Python to run the app
-      if: ${{ !startsWith(inputs.runner-os, 'ubuntu') }}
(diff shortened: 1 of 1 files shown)


rmartin16:
This addresses the issue [seen](https://github.com/beeware/briefcase-linux-appimage-template/pull/46#discussion_r1445447847) in [beeware/briefcase](https://github.com/beeware/briefcase-linux-appimage-template/pull/46). We could also update this to run `setup-python` if Flatpak or AppImage is being explicitly requested....however, that would create a discrepancy in how this runs for those formats between when they are explicitly requested versus no explicit output format.

freakboy3742 (approved):

PR #82, opened 2024-01-12
 Ensure Linux System packages can be installed

rmartin16:
## Changes
- Install Linux system package after it's created in `app-build-verify`

## Notes
- I was inspired to try this after I discovered the openSUSE package couldn't be installed
- I'm not married to adding this....but it seems like a reasonable thing to check

## PR Checklist:

### DIFF
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -228,6 +228,8 @@ jobs:
         xvfb-run briefcase run linux system
         briefcase package linux system --adhoc-sign
 
+        bash -c "sudo apt -y install ./dist/*_0.0.1-1~ubuntu-*_amd64.deb"
+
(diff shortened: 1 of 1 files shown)


freakboy3742 (commented):
Agreed this is probably worth checking - I would *hope* that none of the tools can generate packages that can't be installed... but then, this is Linux... :-)

rmartin16:
> Agreed this is probably worth checking - I would _hope_ that none of the tools can generate packages that can't be installed... but then, this is Linux... :-)

haha...my main concern is with uninstallable dependencies either because we are specifying them in the package incorrectly...or maybe they disappeared in the distro...or changed in a distro upgrade... 🤷🏼

freakboy3742 (approved):
👍

PR #88, opened 2024-02-04
Delete Docker images to free up disk space

rmartin16:
- The GitHub runners have limited disk space and creating Docker images occupies ~8GB of disk space. To ensure later tests, such as Android, can run, this deletes all the Docker images.
- RE: https://github.com/beeware/briefcase/pull/1630#issuecomment-1925456651

## PR Checklist:


freakboy3742 (approved):
Seems reasonable to me. Docker is notorious as a disk-hog, so this isn't all that surprising. 

As a mental note - it would seem to put a limit on adding many more other Linux platforms to the CI list, unless we're going to (a) flush Docker after each platform build, and/or (b) use a matrix build for each Docker target.

PR #96, opened 2024-02-15
print `bash` commands as they are executed

rmartin16:
## Changes
- Use `bash --noprofile --norc -x -e -o pipefail {0}` as the shell for all `run` actions
- I often feel lost in the workflow output; this helps keep track of where you are in the list of commands

## Notes
- My primary concern with this is we no longer inherit the list of `bash` options that GitHub uses by default.
- That said, I think these options are quite stable and unlikely to meaningfully change over time.

## PR Checklist:

### DIFF
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -45,7 +45,7 @@ on:
 
 defaults:
   run:
-    shell: bash
+    shell: bash --noprofile --norc -x -e -o pipefail {0}
 
 env:
   FORCE_COLOR: "1"


freakboy3742 (approved):
I can't say pulling apart the logs has caused me any particular difficulty myself - but then, you've spent a lot more time diagnosing weird CI workflow issues recently, so you might be seeing this more acutely than I am. 

I'm not especially concerned about changes in Bash's arguments, though - if Bash changed an argument of any note, you'd be able to see the nerdrage from orbit :-) I'd be more concerned about missing some neat feature that Github rolls out as a default;  but even then, it's an easy change to roll back (or modify) if that becomes an issue.

PR #97, opened 2024-02-18
 Perform Linux package installs as dry runs 

rmartin16:
Trying to cut down time spent in CI for build apps for Linux.

## PR Checklist:

### DIFF
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -238,7 +238,7 @@ jobs:
         xvfb-run briefcase run linux system
         briefcase package linux system --adhoc-sign
 
-        bash -c "sudo apt -y install ./dist/*_0.0.1-1~ubuntu-*_amd64.deb"
+        bash -c "sudo apt -y install --dry-run ./dist/*_0.0.1-1~ubuntu-*_amd64.deb"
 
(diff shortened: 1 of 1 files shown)


rmartin16:
### Before:
image

### After:
image

sooo, not especially striking...but something, I suppose.

freakboy3742 (commented):
I'm all up for faster CI... but is this really a good idea? A dry run may be faster, but that's because it's *not* doing the thing that we're trying to validate works.

rmartin16:
> I'm all up for faster CI... but is this really a good idea? A dry run may be faster, but that's because it's _not_ doing the thing that we're trying to validate works.

Well, yes and no, I suppose.

The part I really wanted to test by actually installing the produced package was ensuring:
- the package could be properly read by package manager
- the dependencies listed therein are valid according to the package manager

AFAICT, these two steps still happen for a dry run....it just doesn't waste all the additional time of downloading and unpacking those dependencies....which Briefcase really has no bearing on anyway.

freakboy3742 (approved):
That's fair. I guess my concern is that the unpacking process might reveal that a file can't be unpacked, or something like that - but that's probably enough of an edge case that it's not worth worrying about.

PR #99, opened 2024-02-20
Use ghcr.io for Docker image caching

### LINKED ISSUE #98: Cache Docker images for `app-build-verify`
### What is the problem or limitation you are having?

Currently, CI must completely rebuild several images every time it runs. These should be cached and downloaded when needed. This should work for non-`main` branches as well as forks such that even a brand new fork will use a cached image.

### Describe the solution you'd like

Like anything worth doing in Docker, this ain't easy.

Independent of the Docker image build process, you can't just say "take the current state, cache it, and let me restore it when I try to build again"...everything is a bit more tightly integrated. The is primarily because the "state" isn't just the tag assigned at the end; you can't just export that image for that tag, save it somewhere, re-import it later, and expect the `docker build` command to know it doesn't need to redo everything. That image needs additional metadata and potentially even additional layers that wouldn't be normally included for `docker build` to treat it more like a cache.

But even then, the `docker build` command needs to be instructed to use it; therefore, the `docker build` command from Briefcase will almost certainly need to be updated to support arbitrary additional options.

This Docker caching is facilitated via the `--cache-to` and `--cache-from` [options](https://docs.docker.com/build/cache/backends/).

### Option 1: Inline Caching
The simplest way to accomplish this may be ["inline" caching](https://docs.docker.com/build/cache/backends/inline/). When this is configured, Docker treats the output image as the place to cache. So, if you push an image to a registry built with inline caching and pull it later, your `docker build` command can use it.

However, the Briefcase images aren't currently tagged in a way for them to be pushed to a registry....so, there will need to be this dance of associating another tag with the image and then pushing. Conversely, when they are pulled down, they'll need to be re-tagged again.

### Option 2: Direct Registry Caching
Docker supports [caching to and from a registry](https://docs.docker.com/build/cache/backends/registry/) directly....however, they don't support this out of the box for a reason I haven't grasped yet. So, a custom builder needs to be created using the ["docker-container" driver](https://docs.docker.com/build/drivers/docker-container/).

In this way, the `--cache-to` and `--cache-from` options simply need to set to an arbitrary registry image that Docker will treat as a cache.....it's actually pretty snazzy.....but I don't understand the potential consequences of creating and using such a builder with Briefcase yet.

### Describe alternatives you've considered

No caching.

### Additional context

This will serve as a knowledge dump for now....since I realized I won't be able to knock this out very quickly apparently...

rmartin16:
## Changes
- Use a ghcr.io package called "beeware-ci" to cache Docker image builds
- This implementation is inspired by how [caching is managed](https://github.com/indygreg/python-build-standalone/blob/94de081efb139cc48a0cdb9c891374ad9b2cdfe2/.github/workflows/linux.yml#L94) in the python-build-standalone project....so, there's at least _some_ precedent for this approach
- Images for different Briefcase builds are managed by tags on the package
  - This allows CI to manage many independent images while only creating one package in ghcr.io
  - Example `beeware-ci` package: https://github.com/rmartin16/.github-beeware-test/pkgs/container/beeware-ci
- Caching strategy
  - The cache is pulled in from one of three places in this precedence:
    - The package from BeeWare org using the branch for the event
    - The package from BeeWare org using the `main` branch
  - The cache is pushed to the `beeware-ci` package for the BeeWare org
    - So, for a PR, this will be for a tag using the branch name
    - For a merge in to `main`, that will be a tag for `main` branch
    - This caching will likely fail except for the `main` branch because the `GITHUB_TOKEN` for PRs doesn't allow for writing to packages
    - However, that should be fine because the cache can still be pulled using the `main` tag
- Closes #98

## Notes
- Dependent on https://github.com/beeware/briefcase/pull/1661

## PR Checklist:

### DIFF
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -55,6 +55,9 @@ jobs:
     name: Verify App Build
     runs-on: ${{ inputs.runner-os }}
     timeout-minutes: 30
+    permissions:
+      contents: read
+      packages: write
     env:
       SDL_AUDIODRIVER: dummy  # disable audio for SDL
       TOGA_SYSTEM_REQUIRES: libthai-dev libegl1 libcairo2-dev libgirepository1.0-dev gir1.2-gtk-3.0
@@ -240,15 +243,67 @@ jobs:
 
         bash -c "sudo apt -y install --dry-run ./dist/*_0.0.1-1~ubuntu-*_amd64.deb"
 
+    - name: Log in to GitHub Container Registry
+      if: >
+        startsWith(inputs.runner-os, 'ubuntu')
+        && contains(fromJSON('["", "Linux"]'), inputs.target-platform)
+        && contains(fromJSON('["", "system", "appimage"]'), inputs.target-format)
+      uses: docker/login-action@v3.0.0
+      with:
+        registry: ghcr.io
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Set up Docker Buildx
+      if: >
+        startsWith(inputs.runner-os, 'ubuntu')
+        && contains(fromJSON('["", "Linux"]'), inputs.target-platform)
+        && contains(fromJSON('["", "system", "appimage"]'), inputs.target-format)
+      uses: docker/setup-buildx-action@v3.0.0
+
+    - name: Docker Caching Configuration
+      id: docker
+      if: >
+        startsWith(inputs.runner-os, 'ubuntu')
+        && contains(fromJSON('["", "Linux"]'), inputs.target-platform)
+        && contains(fromJSON('["", "system", "appimage"]'), inputs.target-format)
+      env:
+        GITHUB_REF_NAME: ${{ github.ref_name }}
+        DEFAULT_BRANCH: ${{ github.event.repository.default_branch }}
+      run: |
+        PKG_NAME="beeware-ci"
+        REGISTRY_PKG="ghcr.io/$(cut -d/ -f1 <<< "${{ github.repository }}")/${PKG_NAME}"
+        CURRENT_REF="$(sed 's|[^a-z0-9_-]|_|g' <<< "${GITHUB_REF_NAME,,}")"
+        DEFAULT_REF="$(sed 's|[^a-z0-9_-]|_|g' <<< "${DEFAULT_BRANCH,,}")"
+
+        # include all layers, minimize compression time, and ignore package upload issues
+        CACHE_TO_OPTS="mode=max,compression-level=0,ignore-error=true"
+
+        BUILDX_OPTIONS=(
+          # attempt to push to BeeWare's package
+          --Xdocker-build=--cache-to=type=registry,ref=${REGISTRY_PKG}:__PKG_TAG__-${CURRENT_REF},${CACHE_TO_OPTS}
+          # retrieve the cache using the current branch, falling back to the default branch
+          --Xdocker-build=--cache-from=type=registry,ref=${REGISTRY_PKG}:__PKG_TAG__-${CURRENT_REF}
+          --Xdocker-build=--cache-from=type=registry,ref=${REGISTRY_PKG}:__PKG_TAG__-${DEFAULT_REF}
+        )
+        echo "DOCKER_BUILDX_OPTIONS=${BUILDX_OPTIONS[@]}" | tee -a ${GITHUB_ENV}
+
+        # Tag format: app-build-<year>-<month>-<framework>-<python-version>-<format>-<format-differentiators>-<branch>
+        # For example: app-build-24-02-toga-3.11-system-debian-bookworm-main
+        echo "tag-base=app-build-$(date +%Y-%m)-${{ inputs.framework }}-${{ inputs.python-version }}" | tee -a ${GITHUB_OUTPUT}
+
     - name: Build Linux System Project (Debian, Dockerized)
       if: >
         startsWith(inputs.runner-os, 'ubuntu')
         && contains(fromJSON('["", "Linux"]'), inputs.target-platform)
         && contains(fromJSON('["", "system"]'), inputs.target-format)
       working-directory: ${{ steps.create.outputs.project-path }}
+      env:
+        PKG_TAG: ${{ steps.docker.outputs.tag-base }}-system-debian-bookworm
       run: |
         briefcase create linux system --target debian:bookworm \
-          ${{ steps.output-format.outputs.template-override }}
+          ${{ steps.output-format.outputs.template-override }} \
+          ${DOCKER_BUILDX_OPTIONS//__PKG_TAG__/${PKG_TAG}}
         briefcase build linux system --target debian:bookworm
         xvfb-run briefcase run linux system --target debian:bookworm
         briefcase package linux system --target debian:bookworm --adhoc-sign
@@ -262,9 +317,12 @@ jobs:
         && contains(fromJSON('["", "Linux"]'), inputs.target-platform)
         && contains(fromJSON('["", "system"]'), inputs.target-format)
       working-directory: ${{ steps.create.outputs.project-path }}
+      env:
+        PKG_TAG: ${{ steps.docker.outputs.tag-base }}-system-fedora-39
       run: |
         briefcase create linux system --target fedora:39 \
-          ${{ steps.output-format.outputs.template-override }}
+          ${{ steps.output-format.outputs.template-override }} \
+          ${DOCKER_BUILDX_OPTIONS//__PKG_TAG__/${PKG_TAG}}
         briefcase build linux system --target fedora:39
         xvfb-run briefcase run linux system --target fedora:39
         briefcase package linux system --target fedora:39 --adhoc-sign
@@ -278,9 +336,12 @@ jobs:
         && contains(fromJSON('["", "Linux"]'), inputs.target-platform)
         && contains(fromJSON('["", "system"]'), inputs.target-format)
       working-directory: ${{ steps.create.outputs.project-path }}
+      env:
+        PKG_TAG: ${{ steps.docker.outputs.tag-base }}-system-arch
       run: |
         briefcase create linux system --target archlinux:latest \
-          ${{ steps.output-format.outputs.template-override }}
+          ${{ steps.output-format.outputs.template-override }} \
+          ${DOCKER_BUILDX_OPTIONS//__PKG_TAG__/${PKG_TAG}}
         briefcase build linux system --target archlinux:latest
         xvfb-run briefcase run linux system --target archlinux:latest
         briefcase package linux system --target archlinux:latest --adhoc-sign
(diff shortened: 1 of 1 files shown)


rmartin16:
@freakboy3742, now that I have all this working and before I complete the work, do you agree with this approach? Questions or concerns about the implementation and its implications?

Here are the PoC Briefcase [changes](https://github.com/beeware/briefcase/compare/main...rmartin16:briefcase:extra-docker-build-args) as well.

freakboy3742:
The general approach makes sense; two questions:

1. CI times seem to have increased as a result of this caching. The Verify Linux (toga) step in this PR's CI took 13:40; the most recent main CI build took 11:06 for the same step. Is this just a first-time build cache cost, or is there something else going on?
2. This is going to create tagged Docker images for every PR - won't that space grow continuously? What process will purge the image cache?

rmartin16:
> The general approach makes sense; two questions:
> 
> 1. CI times seem to have increased as a result of this caching. The Verify Linux (toga) step in this PR's CI took 13:40; the most recent main CI build took 11:06 for the same step. Is this just a first-time build cache cost, or is there something else going on?

Nothing is able to use caching so far; the caches haven't been seeded yet. In my testing, I was seeing as much as 5 - 7 minutes shaved off each using caching. Although, this seems to be particularly contingent on the Fedora package CDN not dropping to 150kb/s for the dry-run install...
 
> 2. This is going to create tagged Docker images for every PR - won't that space grow continuously? What process will purge the image cache?

Nothing; however, my expectation of an image registry means this shouldn't really be necessary. AFAIU, a tag is mostly a manifest of image layer hashes. So, just because a new tag is created does not mean that wholly new images are being uploaded to the registry. For the vast majority of CI runs, they will use and create the exact same image layer hashes that already exist in the registry. So, all that really grows over time is the list of tags.

Additionally, I'm still debating if we need to support uploading packages outside of the BeeWare namespace. I have a little more testing to do but I think that `GITHUB_TOKEN` will be valid for members of the BeeWare org to push packages....and non-members attempt to push will simply fail. That would simplify this a bit....especially since the vast majority of PRs won't have any impact on what's in these images.

To that end, @freakboy3742, can you check if the `GITHUB_TOKEN` for the BeeWare has write privileges enabled?

https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/#setting-the-default-permissions-for-the-organization-or-repository

And if we're not really comfortable giving the `GITHUB_TOKEN` write permissions, then we can use a specific token....but that will definitely prevent uploading packages elsewhere. That said, only workflows that explicitly have a `permissions` section will be able to use the write permission anyway.

freakboy3742:
> > The general approach makes sense; two questions:
> > 
> > 1. CI times seem to have increased as a result of this caching. The Verify Linux (toga) step in this PR's CI took 13:40; the most recent main CI build took 11:06 for the same step. Is this just a first-time build cache cost, or is there something else going on?
> 
> Nothing is able to use caching so far; the caches haven't been seeded yet. 

👍 
 
> > 2. This is going to create tagged Docker images for every PR - won't that space grow continuously? What process will purge the image cache?
> 
> Nothing; however, my expectation of an image registry means this shouldn't really be necessary. AFAIU, a tag is mostly a manifest of image layer hashes. So, just because a new tag is created does not mean that wholly new images are being uploaded to the registry. For the vast majority of CI runs, they will use and create the exact same image layer hashes that already exist in the registry. So, all that really grows over time is the list of tags.

Sure, but the *contents* of the hash still needs to be stored - and in our case, that contents will include all the packages that have been installed, which will change every time the upstream repo pushes an update to each individual package. That's potentially a couple of hundred meg for each hash, isn't it?
 
> To that end, @freakboy3742, can you check if the `GITHUB_TOKEN` for the BeeWare org has write privileges enabled?

I can confirm write permissions are enabled for the Beeware org:

Screenshot 2024-02-21 at 10 25 34 am

rmartin16:
> > > 3\. This is going to create tagged Docker images for every PR - won't that space grow continuously? What process will purge the image cache?
> > 
> > 
> > Nothing; however, my expectation of an image registry means this shouldn't really be necessary. AFAIU, a tag is mostly a manifest of image layer hashes. So, just because a new tag is created does not mean that wholly new images are being uploaded to the registry. For the vast majority of CI runs, they will use and create the exact same image layer hashes that already exist in the registry. So, all that really grows over time is the list of tags.
> 
> Sure, but the _contents_ of the hash still needs to be stored - and in our case, that contents will include all the packages that have been installed, which will change every time the upstream repo pushes an update to each individual package. That's potentially a couple of hundred meg for each hash, isn't it?

The Docker build process will use the cached layer instead of re-running the step and creating a new layer....so, even if Debian uploads new packages, they will not be installed in to the Docker image in CI.

This is actually quite similar to the problem we had with the Briefcase cache and we added the year and month to the cache key to prevent it becoming too stale. We'll probably need to do the same thing here.

Furthermore, though, my testing in #100 shows that even PRs from BeeWare org members will not be able to upload to packages; so, for the most part, the only tags for the package will be from the `main` branch....which should be fine in practice, I think.

Is there a concern about uploading new tags that I'm missing, though? Want to make sure we're not talking past each other.

rmartin16:
I updated the caching strategy to only attempt to push and pull from the BeeWare org package namespace. The use-case for pushing to a user's package namespace was mostly for forks that aren't committing back upstream...but I don't think anybody is really doing that.

Additionally, since the `GITHUB_TOKEN` doesn't have write permissions for PRs from forks apparently, it will really only be the `main` branch that ever actually pushes to the package without erroring.

freakboy3742:
> > > > 3. This is going to create tagged Docker images for every PR - won't that space grow continuously? What process will purge the image cache?
...
> Is there a concern about uploading new tags that I'm missing, though? Want to make sure we're not talking past each other.

I can only assume I'm misunderstanding what is being cached/optimized here. 

My understanding is that the process works broadly like this:

As a user, you define a Dockerfile. This describes a starting image (ubuntu/latest), and series of instructions for building on that image. 

Each instruction forms a new hash, which is uniquely addressable, so if you've got instructions A B C, and you modify the instructions to A B D, the first two can be reused. 

The hash of each layer is a unique identifier, but for convenience, you can associate a tag with a hash - so you can refer to the `1deadbeefcafe` hash `myproject/latest`. 

The hash is just an identifier. On disk, actual files are being downloaded and stored. However, it's efficient "delta-based" storage - if you have a big base image hash A, and add 1 tiny script file in B, the incremental storage from storing both the base image and the script file is A+B, not 2A + B.

The hash is also a *unique* identifier. If *I* define a docker file that adds an "echo helloworld" instruction on top of ubuntu/latest, and you do the same, they'll be the same hash.

This becomes especially useful when repositories get added to the mix. Every local Docker install has a repository, which is used as a local cache. When you build an image based on ubuntu/latest, the first step is downloading that image from an upstream repo, and caching it locally. When I write my Dockerfile, the layers, keyed by hash, are stored in my local repo.

If I have a Dockerfile that is particularly interesting, I can then publish that hash to an upstream repo (like ghcr.io), using a human-readable tag. So - if I build an image based on ubuntu/latest that adds layers A B C, I can publish that as rkm/latest. This will push the hashes for A B and C, and the deltas of content for A B and C, plus the rkm/latest label. Another user can then pull the rkm/latest image from ghcr.io and use it (and build on it, if they want). 

The useful feature for our purposes here is that if I have published rkm/latest, and a second user pulls rkm/latest into their local repo, and then they try to run a Dockerfile that defines an image rkm/latest + A B C configuration, that will evaluate to the same hash, and the image that has been pulled from ghcr.io can be used as is, rather than needing to go through the processes of actually running the commands to evaluate A, B and C (which, in the case of an apt update, could be time consuming). As long as the time required to download the layers for A B C is less than the time to *build* A B C from scratch (and it almost certainly will be for any apt install), you've got a net win.

So - based on that understanding: My concern is the storage associated with these tags. We're not talking about a single image - we're talking about publishing A B C, A B D, A B E, ... with each PR essentially producing a different hash to be pushed to ghcr.io; and while the storage is a delta (in that we won't be reproducing the storage of the whole of ubuntu with each image), we *will* be storing the C, D, E,... deltas - and in our case, those can be a whole bunch of apt installs of entire compiler suites - potentially a hundred megabytes or more each. 

That isn't a problem in itself.... until we run the clock forward, and we've been pushing hundred megabyte layer deltas on every pull request for a year. In my own experience, I need to purge my local docker cache *regularly* - `docker system prune -a` releasing dozens of gigabytes is disturbingly common. In the local case, this can be mitigated by remembering to use `docker run -rm` so that ephemeral images aren't retained; but if we're actually pushing images to a central repo, I would have thought storage on that repo would *eventually* become a problem, and we would need to consider how to purge images that are a month old and won't ever be used again.

Have I missed something in my understanding?

rmartin16:
> Have I missed something in my understanding?

AFAICT, you're understanding is correct...however, I think the application of it to our situation and therefore the implications of that application could be over-inflated.

> we're talking about publishing A B C, A B D, A B E, ... with each PR essentially producing a different hash to be pushed to ghcr.io

As you said above:

> The hash is also a unique identifier. If I define a docker file that adds an "echo helloworld" instruction on top of ubuntu/latest, and you do the same, they'll be the same hash.

Therefore, all the PRs calculate the same hashes and skip the step altogether...so, the PRs don't need to upload anything except for the tag because all the hashes already exist in ghcr.io; and if they don't, this push will seed the cache.

To this end, I added the year and month to the tag so the tags may include new image layers once a month....unless of course the actual Dockerfiles change since that would also create new images layers.

>  would have thought storage on that repo would eventually become a problem, and we would need to consider how to purge images that are a month old and won't ever be used again.

It may; GitHub is allowing open source projects to use seemingly unlimited storage right now. From personal experience, I haven't had any problem; for instance, I am regularly producing hundreds of new tags (and quite likely with new hashes) for my [docker-qbittorent](https://github.com/rmartin16/docker-qbittorrent/pkgs/container/qbittorrent-nox/versions) project.

I guess from my standpoint, if GitHub wants to change this storage situation, then they'll probably need to start pruning old image layers somehow.....or stop accepting new pushes. Either way, I expect some sort of notification with the time to react.

Alternatively, I'm not against something to delete the package regularly. I do find this line in the docs a bit interesting, though:

> If the package is public and the package version has more than 5,000 downloads, you cannot delete the package version. In this scenario, contact GitHub support for further assistance.

https://docs.github.com/en/rest/packages/packages?apiVersion=2022-11-28#delete-a-package-version-for-the-authenticated-user

freakboy3742:
> > The hash is also a unique identifier. If I define a docker file that adds an "echo helloworld" instruction on top of ubuntu/latest, and you do the same, they'll be the same hash.
> 
> Therefore, all the PRs calculate the same hashes and skip the step altogether...so, the PRs don't need to upload anything except for the tag because all the hashes already exist in ghcr.io; and if they don't, this push will seed the cache.
> 
> To this end, I added the year and month to the tag so the tags may include new image layers once a month....unless of course the actual Dockerfiles change since that would also create new images layers.

Ok - I think that's the source of my confusion. I was forseeing a larger proliferation of images than we're going to see in practice. I was forgetting that the image for any given BeeWare CI build is essentially identical for all PR runs, with the only real discrepancies happening when either the Dockerfile changes, or a specific PR adds a new dependency to the system/system_requires list - both of which are fairly infrequent changes; and regardless, would be a delta on top of the shared *base* hash which does the initial system update and Python install.
 
> > would have thought storage on that repo would eventually become a problem, and we would need to consider how to purge images that are a month old and won't ever be used again.
> 
> It may; GitHub is allowing open source projects to use seemingly unlimited storage right now. 

With my grey hat on, I can see this as a way to get unlimited free archival disk storage... 

> Alternatively, I'm not against something to delete the package regularly. 

The problem I was forseeing is definitely a lot less acute given that we're going to be pushing a lot fewer images than I originally thought. If there's no in-practice issue right now, then I don't think we need to invent one. As long as we've got a reasonable potential strategy for purging if/when they introduce a limit, then I'm happy. Since this is entirely a cache/optimization approach, I guess "nuke the whole thing from orbit" is an entirely reasonable option.

> I do find this line in the docs a bit interesting, though:
> 
> > If the package is public and the package version has more than 5,000 downloads, you cannot delete the package version. In this scenario, contact GitHub support for further assistance.
> 
> https://docs.github.com/en/rest/packages/packages?apiVersion=2022-11-28#delete-a-package-version-for-the-authenticated-user

I'm going to guess this is protection against a left-pad style rug pull. Someone builds an image that lots of people integrate into their workflows, then yanks the image  (either intentionally or accidentally), and everyone downstream has to scramble to fix their projects.

freakboy3742:
Just giving this a final lookover- [is this 403 concerning?](https://github.com/beeware/.github/actions/runs/8013325227/job/21891324791?pr=99#step:21:1803)

rmartin16:
> Just giving this a final lookover- [is this 403 concerning?](https://github.com/beeware/.github/actions/runs/8013325227/job/21891324791?pr=99#step:21:1803)

It's an outcome of the `GITHUB_TOKEN` [permissions](https://github.com/beeware/.github/actions/runs/8013325227/job/21891324791?pr=99#step:1:16) for a PR from a fork:
```
GITHUB_TOKEN Permissions
  Contents: read
  Metadata: read
  Packages: read
```
So, CI will only be able to actually push a tag on `main` branch or for PRs on the repo itself, like [CI](https://github.com/beeware/.github/actions/runs/7983571675/job/21798960352#step:21:135) for #100.

freakboy3742 (approved):
As always, the twisty passages of Github permissions and "CI testing CI configurations" makes this a bit of a headache to confirm, but I think this all makes sense.

PR #106, opened 2024-03-06
Add workflow to bump unmanaged dependency versions

rmartin16:
## Changes
- Adds a workflow to keep pinned dependencies in `pyproject.toml` and `tox.ini` updated
- Also adds the `install_requirement.py` script from Briefcase so other repos can use it

## PR Checklist:


freakboy3742 (approved):
I'll resist the temptation to say "But what about the dependencies of the dependency update script?"... :-)

Looks solid, and you can't argue with the success of the example runs you've shown.

PR #137, opened 2024-05-24
Implement `hynek/build-and-inspect-python-package` in `python-package-create.yml`

rmartin16:
## Changes
- Outsource's package building to [hynek/build-and-inspect-python-package](https://github.com/hynek/build-and-inspect-python-package)
- It looks like Hynek is fostering some decent oversight on this workflow and has the necessary community participation to add existing and upcoming attestation and provenance metadata for packages

## Implementations
- https://github.com/beeware/beeware/pull/375
- https://github.com/beeware/briefcase/pull/1843
- https://github.com/beeware/gbulb/pull/148
- https://github.com/beeware/rubicon-objc/pull/474
- https://github.com/beeware/toga/pull/2604
- https://github.com/beeware/toga-chart/pull/131
- https://github.com/beeware/travertino/pull/180

## PR Checklist:


rmartin16:
This change replaces the current package building with [hynek/build-and-inspect-python-package](https://github.com/hynek/build-and-inspect-python-package); it is quite similar to the current process:
- `python -m build` to create the sdist and wheel
- `python -m twine check` verification
- `check-wheel-contents` verification (new)
- [GitHub attestation](https://github.blog/2024-05-02-introducing-artifact-attestations-now-in-public-beta/) (new)

So, along with our current checks, we gain `check-wheel-contents` verification and attestation. See existing examples of attestation in practice at [structlog](https://github.com/hynek/structlog/attestations) or [pytest-mock](https://github.com/pytest-dev/pytest-mock/attestations).

I'm also expecting this Action to add support for the upcoming support within PyPI for declaring provenance....which we should be able to get more or less for free.

Along with https://github.com/beeware/briefcase/pull/1839 and https://github.com/beeware/beeware/pull/371, I'll need to update the repos below to properly use this new version of `python-package-create.yml` as well as update their `tox.ini` to run `check-wheel-contents` for `tox -e package`.
- beeware
- briefcase
- gbulb
- rubicon-objc
- toga
- toga-android
- toga-chart
- travertino

I'll pause here, though, to ensure there aren't general objections to using Hynek's GitHub Action for this purpose.

rmartin16:
Also, here's an example of the attestation working for BeeWare: https://github.com/beeware/.github/attestations

This PR (as well as any PR from a fork) cannot create attestations; so, I created these initial ones using https://github.com/beeware/.github/pull/138. This means we can't enable attestations across the board; instead, the `release.yml` workflows in each repo will need to pass a flag to `ci.yml` to tell `python-package-create.yml` to create an attestation.

freakboy3742:
tl;dr - no objection to adopting this. As you've described - Hynek is well known in the Python ecosystem, and he sweats the little details when it comes to issues like CI and packaging; so if we can leverage his work, all the better.

.github/workflows/ci.yml
```diff
@@ -161,38 +161,48 @@ jobs:
...
+          repo: briefcase
+          build-subdir: automation
+        - name: toga-core
+          repo: toga
           build-subdir: core
+        - name: toga-android
+          repo: toga
+          build-subdir: android
+
+  test-package-briefcase:
```
rmartin16, line 189: I had to separate building the Briefcase package out because the `artifact-name` output from the `python-package-create.yml` becomes unstable when it is used in a matrix. This is under-designed in the GitHub Actions system and it just takes on the value from whichever job set it last. I found several proposals to fix this but GitHub hasn't gotten around to this. There are a few third-party workarounds to this but they seem very heavy-handed for our use-case.

Furthermore, my whole goal with outputting the Artifact's name was to abstract the packaging process from its downstream use; for the most part, I've achieved that. However, parts of the implementation are assuming that all related packages share a common base name. So, for Toga, it's assumed all of the Artifact's names start with `Packages-toga`....a more complicated implementation could probably workaround this but whatever....

.github/workflows/app-build-verify.yml
```diff
@@ -127,7 +131,8 @@ jobs:
       if: endsWith(inputs.repository, 'briefcase')
       uses: actions/download-artifact@v4.1.7
       with:
-        name: packages-briefcase
+        pattern: ${{ format('{0}*', inputs.briefcase-artifact-name) }}
+        merge-multiple: true
```
freakboy3742, line 135: It's not clear to me why this needs to be parameterised. AFAICT, it's only used by the briefcase repo, and will only ever be set to the name of the briefcase package. What do we gain by parameterising the name of the briefcase artefact?

.github/workflows/app-create-verify.yml
```diff
@@ -79,18 +87,18 @@ jobs:
 
     - name: Get Briefcase Package
       # Briefcase will build and package itself in a previous step in its CI
-      if: endsWith(github.repository, 'briefcase')
+      if: endsWith(inputs.repository, 'briefcase')
       uses: actions/download-artifact@v4.1.7
       with:
-        name: packages-briefcase
+        name: ${{ inputs.briefcase-artifact-name }}
```
freakboy3742, line 93: As with app-build-verify - what do we gain by parameterisation here?

.github/workflows/ci.yml
```diff
@@ -161,38 +161,48 @@ jobs:
...
-        repo: [ briefcase, gbulb, rubicon-objc, toga ]
+        name:
+        - beeware
+        - briefcase-automation
+        - gbulb
+        - rubicon-objc
+        - toga-core
+        - toga-android
+        - toga-chart
+        - travertino
```
freakboy3742, line 177: What are the extra test cases validating here? briefcase-automation and toga-android make sense - they're both variations on sub packages; but it's not clear to me what validating beeware, toga-chart and travertino is adding over the base testing of gbulb as a "simple top-level package"

If there's explicit use cases being tested, it would be worth documenting them so it's clear why these projects are part of the CI checks.

.github/workflows/python-package-create.yml
```diff
@@ -1,98 +1,80 @@
...
+        value: ${{ jobs.package.outputs.artifact-basename }}
 
 env:
   FORCE_COLOR: "1"
 
 jobs:
   package:
     name: Create Python Package
-    runs-on: ${{ inputs.runner-os }}
+    runs-on: ubuntu-latest
```
freakboy3742, line 41: Does this mean the rubicon-objc package is being built on Ubuntu? I guess raw packaging shouldn't matter... it just seems a bit odd.

freakboy3742 (changes requested):
This mostly makes sense; a couple of questions about possible over-parameterization, and testing/runner usage strategy.

I've reviewed the repos using these changes; two comments that came from that review:

1. Is preserving the tox workflow for packaging is worthwhile? AFAICT we're not actually using it for "real" packaging, so there's a risk that it could diverge from reality; to that end, I wonder if we'd be better served removing that target from the toxfile to remove the suggestion that it is meaningful.
2. Now that `install_requirement.py` is available as a resource in `.github`, do we still need the copy in the `briefcase` repo? Can we rationalise that into the single version in `.github`?

.github/workflows/app-build-verify.yml
rmartin16: Mostly because the name of the artifact is arbitrary. We can hardcode it here but that just means this workflow is dependent on however the package is uploaded as an artifact. Is that what you mean?

freakboy3742: > Mostly because the name of the artifact is arbitrary. 

Well... yes and no. Yes. the artefact *could* be named anything... but briefcase only ever uses one name, and this is only ever used by Briefcase. Unless we're anticipating Briefcase regularly changing its artefact name, I'm not sure what we gain by parameterization of the briefcase artefact name.

.github/workflows/ci.yml
rmartin16: Oh, I just added every repo currently using `python-package-create.yml`. I can trim it back down....although this workflow runs quite fast now so it definitely doesn't hit CI runtime.

.github/workflows/app-build-verify.yml
freakboy3742: (to be clear - this is a case where parameterisation definitely isn't *wrong* - I'm just wary that there's already a lot of moving parts and interactions between obscure parts of the Actions API, so I'm keen to keep this as simple as humanly possible)

.github/workflows/python-package-create.yml
rmartin16: Hynek's action is only documented to run on Linux. Presently, it fails on the `tree` command if you don't take steps to install it beforehand.

.github/workflows/app-build-verify.yml
rmartin16: That's fair; if the artifact name changes, it'll be obvious to change it here.

rmartin16:
> 1. Is preserving the tox workflow for packaging is worthwhile? AFAICT we're not actually using it for "real" packaging, so there's a risk that it could diverge from reality; to that end, I wonder if we'd be better served removing that target from the toxfile to remove the suggestion that it is meaningful.

I think I could go either way; this is a bit similar to pre-commit insofar as CI doesn't use Tox for pre-commit either....although, pre-commit in Tox is a definitely more useful than packaging is. I think the likelihood for significant divergence is probably low....and the workflow will be useful if packaging issues arise....although, running the commands manually isn't a big deal. So...I'm fine getting rid of it if you see the balance going more that way.,

> 2. Now that `install_requirement.py` is available as a resource in `.github`, do we still need the copy in the `briefcase` repo? Can we rationalise that into the single version in `.github`?

This is something I've been wanting to reconcile. The primary issue is Briefcase is using this script in `tox.ini`. The obvious solution is just to stop using it in Tox and go back to just installing the `[dev]` dependencies. This negates, at least some of, the original intention of the script....but the script is arguably much more valuable in CI thanks to Tox caching environments locally. Ideally, I think, we'd be able to use the script both in Tox and CI for all repos....but I don't see an easy way to do that if the script only lives in `beeware/.github`....maybe by turning `beeware/.github` into an installable Python package....

At any rate, I think it does makes sense to remove `install_requirement.py` from `beeware/briefcase` and therefore from `tox.ini` as well. Furthermore, though, I should probably create a composite GitHub Action that runs `install_requirement.py` instead of copy pasting all this boilerplate code in to all the repos' CI.

rmartin16:
I created https://github.com/beeware/.github/pull/139/files to introduce a GitHub Action to run `install_requirement.py` in arbitrary repos' CI. I'll rebase this once that is merged.

freakboy3742:
> I think I could go either way; this is a bit similar to pre-commit insofar as CI doesn't use Tox for pre-commit either....although, pre-commit in Tox is a definitely more useful than packaging is. I think the likelihood for significant divergence is probably low....and the workflow will be useful if packaging issues arise....although, running the commands manually isn't a big deal. So...I'm fine getting rid of it if you see the balance going more that way.,

I think the big difference is that pre-commit is something that *everyone* needs to run, but packaging is something that, ideally, *nobody* should be running. Manual packaging is a definite edge case; if we end up in that hole, I'd suggest we'd be better served re-running the workflow and trashing a release than trying to fix it manually. 

So - since we're doing all this re-org across all the projects, my inclination is to trash the tox configs while we're at it.

> > 2. Now that `install_requirement.py` is available as a resource in `.github`, do we still need the copy in the `briefcase` repo? Can we rationalise that into the single version in `.github`?
> 
> This is something I've been wanting to reconcile. The primary issue is Briefcase is using this script in `tox.ini`. 

Ah - I forgot that tox is using this one as well. 

> maybe by turning `beeware/.github` into an installable Python package....

... I mean ... `install_requirements` is a pretty well constrained tool... it wouldn't be the worst thing to turn into a standalone package...

rmartin16:
All right...
- removed `install_requirement.py` from `beeware/briefcase`
- deferred creating an installable package for `install_requirement.py` but that might be reasonable later...
- removed `tox -e package` from the repos

Now that `install_requirement.py` is a GitHub Action, it probably makes sense to update `pre-commit-run.yml` and `towncrier-run.yml` to use it. Of course, though, that will require another round of updates for all repos using them.

rmartin16:
> Now that `install_requirement.py` is a GitHub Action, it probably makes sense to update `pre-commit-run.yml` and `towncrier-run.yml` to use it. Of course, though, that will require another round of updates for all repos using them.

I looked at bit at these today; while they may benefit from implementing the `install-requirement` action, they will need to maintain the existing functionality since many repos don't have a `pyproject.toml`.

freakboy3742 (approved):
👍 

I presume the process from here is that this needs to be landed, then one last CI pass on all the downstream repos for confirmation, then land those repos?

rmartin16:
> I presume the process from here is that this needs to be landed, then one last CI pass on all the downstream repos for confirmation, then land those repos?

Yep; this one first and then the others. It may be sufficient to just re-run the failed CI run....but with how GitHub caches, I may need to re-push the commit.

PR #139, opened 2024-05-28
Introduce GitHub Action to install project requirements

rmartin16:
## Changes
- Introduces composite action to install specific requirements from a project's `pyproject.toml`

## PR Checklist:

### DIFF
--- a/.github/actions/install-requirement/action.yml
+++ b/.github/actions/install-requirement/action.yml
@@ -47,5 +47,5 @@ runs:
       shell: bash
       run: |
         python -m install_requirement ${{ inputs.requirements }} \
-          --extra "${{ inputs.extra }}" \
+          ${{ inputs.extra && format('--extra "{0}"', inputs.extra) || '' }} \
           --project-root "$(printf -- "%q" "${{ steps.paths.outputs.project-root }}")"
--- a/.github/actions/install-requirement/action.yml
+++ b/.github/actions/install-requirement/action.yml
@@ -1,0 +1,51 @@
+name: Install Project Requirements
+description: Install requirements from a PEP 517 project
+
+inputs:
+  requirements:
+    description: "Whitespace-delimited list of project requirements to install."
+    required: true
+  project-root:
+    description: "The filesystem path to the root of the project to install from; defaults to github.workspace."
+    default: ${{ github.workspace }}
+  extra:
+    description: "Name of the optional dependencies marker; e.g. dev."
+    required: false
+    default: ""
+  workflow-repo:
+    # Only needed for PRs in other repos wanting to test new workflow changes before they are merged.
+    # These inputs should not be specified by another repo on their main branch.
+    description: "The repo to use to run additional workflows and actions."
+    default: "beeware/.github"
+  workflow-repo-ref:
+    description: "The repo ref to use to run additional workflows and actions."
+    default: ""
+
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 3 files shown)


.github/actions/install-requirement/action.yml
```diff
@@ -0,0 +1,51 @@
+name: Install Project Requirements
+description: Install requirements from a PEP 517 project
+
+inputs:
+  requirements:
+    description: "Whitespace-delimited list of project requirements to install."
```
freakboy3742, line 6: Given it's being specified as a list, it would be nice if this was an actual list/array, rather than needing to do the whitespace-interpretation dance. AFAICT all that should be required is a `{{ join(inputs.requirements, " " }}` where it's used?

.github/actions/install-requirement/action.yml
```diff
@@ -0,0 +1,51 @@
...
+    required: true
+  project-root:
+    description: "The filesystem path to the root of the project to install from; defaults to github.workspace."
+    default: ${{ github.workspace }}
+  extra:
+    description: "Name of the optional dependencies marker; e.g. dev."
+    required: false
+    default: ""
```
freakboy3742, line 14: Checking that I'm reading this right - if you *don't* specify extra as an argument, this manifests as `--extra ""` in the final arguments, right? Would it be worth gating the `--extra` argument with an `if inputs.extra`?

freakboy3742 (changes requested):
All looks fairly straightforward; two suggestions around argument handling that might help avoid surprises in the future.

.github/actions/install-requirement/action.yml
rmartin16: The default value of `extra` within the `install_requirement.py` script is an empty string. So, passing `--empty` an empty string at the command-line is equivalent to not specifying `--extra` at all. With that, we _can_ control specifying `--extra` at the command-line but it's effectively duplicating effort the script already handles.

.github/actions/install-requirement/action.yml
rmartin16: AFAIR, GitHub doesn't allow us to use complex parameters in to these types of Actions; all inputs are strings. (On a related note, that's also why I have to pass `"true"` to Hynek's Action for attestation instead of just `true`.)

.github/actions/install-requirement/action.yml
rmartin16: That said...

```diff
         python -m install_requirement ${{ inputs.requirements }} \
-          --extra "${{ inputs.extra }}" \
+          ${{ inputs.extra && format('--extra {0}', inputs.extra) || '' }} \
           --project-root "$(printf -- "%q" "${{ steps.paths.outputs.project-root }}")"
```

.github/actions/install-requirement/action.yml
rmartin16: I think this is a limitation of the entire framework for actions regardless of the type.

For instance, consider `actions/setup-python`:
```yaml
steps:
- uses: actions/checkout@v4
- uses: actions/setup-python@v5
  with:
    python-version: '3.10'
    cache: 'pip'
    cache-dependency-path: |
      **/setup.cfg
      **/requirements*.txt
```
They use `|` for `cache-dependency-path` to make it feel like a list.....but that's just fancy YAML for space-delimited list.

So, I think array or list support is limited to YAML that the core GitHub Actions infrastructure processes.

freakboy3742 (approved):
👍

PR #135, opened 2024-05-20
Re-enable OpenSUSE install test

rmartin16:
## Changes
- Closes beeware/briefcase#1802

This reverts commit 66a27a9779b2f57e8adcec17dbd0eac91d006d24.

## PR Checklist:

- [x] All new features have been tested
- [x] All new features have been documented
- [X] I have read the **CONTRIBUTING.md** file
- [X] I will abide by the code of conduct

PR #150, opened 2024-07-11
Deprecate PySide6 AppImage testing.

freakboy3742:
Fixes beeware/briefcase#1908.

AppImage support is explicitly best-effort at this point; since PySide6 support is causing problems, it gets put into the "ignored" pile.

## PR Checklist:

### DIFF
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -378,7 +378,9 @@ jobs:
       # push) is fundamentally incompatible with using an old base image. As of today,
       # it's impossible to install Toga on *any* manylinux image (
       # https://gitlab.gnome.org/GNOME/pygobject/-/issues/590); PySide2 can't be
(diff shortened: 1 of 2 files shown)


rmartin16 (approved):
Looks good. I was a little surprised the most recent [CI run](https://github.com/beeware/.github/pull/135) for briefcase-linux-appimage-template did _not_ fail. I guess that may imply the root cause is not PySide6 itself...

freakboy3742:
> Looks good. I was a little surprised the most recent [CI run](https://github.com/beeware/briefcase-linux-appimage-template/actions/runs/9434427976/job/25986551936) for briefcase-linux-appimage-template did _not_ fail. I guess that may imply the root cause is not PySide6 itself...

My guess would be that the culprit is Applmage/LinuxDeploy. They don't tag their releases, and use a continuous release strategy, so it's really easy for bugs to slip in and have no evidence for it. The Qt plugin is also compiled, so it's another source of segfaulting errors.

PR #159, opened 2024-08-09
Refactor PR Body regexing into independent GH Action

rmartin16:
## Changes
- Introduces `pr-body-parse` composite GitHub Action to parse key/value from body of PR triggering CI

## PR Checklist:

### DIFF
--- a/.github/actions/install-briefcase/action.yml
+++ b/.github/actions/install-briefcase/action.yml
@@ -55,8 +55,8 @@ runs:
       id: briefcase
       shell: bash
       run: |
-        REPO="${{ steps.briefcase-repo-override.outputs.extracted-value || inputs.briefcase-url }}"
-        REF="${{ steps.briefcase-ref-override.outputs.extracted-value || inputs.briefcase-override-version }}"
+        REPO="${{ steps.briefcase-repo-override.outputs.value || inputs.briefcase-url }}"
+        REF="${{ steps.briefcase-ref-override.outputs.value || inputs.briefcase-override-version }}"
 
         printf -- "repo=%s\n" "${REPO}" | tee -a ${GITHUB_OUTPUT}
         printf -- "ref=%s\n" "${REF}" | tee -a ${GITHUB_OUTPUT}
--- a/.github/actions/install-briefcase/action.yml
+++ b/.github/actions/install-briefcase/action.yml
@@ -32,38 +32,38 @@ outputs:
 runs:
   using: composite
   steps:
-    - name: Use Specified Briefcase Version
-      id: briefcase-override
-      env:
-        GITHUB_TOKEN: ${{ github.token }}
+    - name: Checkout beeware/.github
+      uses: actions/checkout@v4.1.7
+      with:
+        path: beeware-install-briefcase-.github
+
+    - name: Check for Briefcase Repo Override
+      id: briefcase-repo-override
+      uses: ./beeware-install-briefcase-.github/.github/actions/pr-body-parse
+      with:
+        key-regex: 'BRIEFCASE[-_]*REPO'
+        testing-pr-body: ${{ inputs.testing-pr-body }}
+
+    - name: Check for Briefcase Ref Override
+      id: briefcase-ref-override
+      uses: ./beeware-install-briefcase-.github/.github/actions/pr-body-parse
+      with:
+        key-regex: 'BRIEFCASE[-_]*REF'
+        testing-pr-body: ${{ inputs.testing-pr-body }}
+
+    - name: Declare Briefcase Source
+      id: briefcase
       shell: bash
       run: |
-        # Check Body of PR for Briefcase version to use
-        # (only PRs will have a value for github.event.pull_request.number)
-        PR_BODY="${{ inputs.testing-pr-body }}"
-        if [[ -z "${PR_BODY}" && -n "${{ github.event.pull_request.number }}" ]]; then
-          PR_BODY=$(gh api repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }} --jq '.body')
-        fi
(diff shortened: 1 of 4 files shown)


rmartin16:
Everything takes longer than you expect...

Couple of notes:
- Both `app-create` and `install-briefcase` check out their own versions of `beeware/.github`
  - This is to ensure there is a reliable location to run `pr-body-parse` from without requiring callers to checkout `beeware/.github` to a specific directory
  - Alternatively, `pr-body-parse` could be run from the `main` branch....but that makes updating the Action harder in the future....and we've also already spent a lot of time talking about this more general situation...
- Handling for `github.token`
  - I'm assuming that the GitHub Token will be handled the same as before....at least insofar as it seemed to be working correctly before and it should still be now
- `printf` shananigans
  - Anywhere the value from the PR body might be printed, I tried to ensure `printf` was used....just to prevent any chance of command injection...

.github/actions/pr-body-parse/action.yml
```diff
@@ -0,0 +1,37 @@
...
+  key-regex:
+    description: "The regex for the key identifying the value to return, e.g. BRIEFCASE[-_]*REPO"
+    required: true
+  testing-pr-body:
+    description: "Override value for body of PR; only for testing."
+    required: false
+
+outputs:
+  extracted-value:
+    value: ${{ steps.parse.outputs.extracted-value }}
```
freakboy3742, line 14: A minor nitpick - but if we're only returning one value, is there any reason to *not* call it `value`, rather than `extracted-value`? These keys are already pretty long, and something like `${{ steps.briefcase-repo-override.outputs.value }}` reads very slightly better.

.github/actions/install-briefcase/action.yml
```diff
@@ -136,8 +136,8 @@ runs:
       id: install
       shell: bash
       run: |
-        REPO="${{ steps.briefcase-override.outputs.repo || inputs.briefcase-url }}"
-        REF="${{ steps.briefcase-override.outputs.ref || steps.briefcase-derived.outputs.version }}"
+        REPO="${{ steps.briefcase.outputs.repo }}"
+        REF="${{ steps.briefcase.outputs.ref || steps.briefcase-derived.outputs.version }}"
```
freakboy3742, line 140: Not sure I follow the inconsistency here - why is the default value still needed for REF, but not for REPO?

freakboy3742 (changes requested):
I guess we *could* have other uses for PR parsing... although I'm having difficulty thinking what other variables we might need to extract. 

Conceptually, I'm on board with the idea of factoring out common behaviours. My concern is that Actions syntax is so verbose that a refactor like this, that sounds like a good idea on paper, ends up with *more* code to yield a "simplification". I'm mildly concerned that the config is sufficiently complex that it's not necessarily obvious how to drive it; adding more complexity is potentially only going to make that worse.

If we do go ahead - a couple of clarifications and suggestions inline.

rmartin16:
> Conceptually, I'm on board with the idea of factoring out common behaviours. My concern is that Actions syntax is so verbose that a refactor like this, that sounds like a good idea on paper, ends up with _more_ code to yield a "simplification". I'm mildly concerned that the config is sufficiently complex that it's not necessarily obvious how to drive it; adding more complexity is potentially only going to make that worse.

Yeah...I am also a bit disappointed with its verbosity. Since, though, I've been thinking that given our use-case is always running two regexes to get the git repo and ref, I could just allow the Action to accept two regexes and output two corresponding values (although, since we can't use YAML lists it would probably require `regex-1` & `regex-2` for inputs and `value-1` and `value-2` for outputs).

.github/actions/install-briefcase/action.yml
rmartin16: Previously, `steps.briefcase-override.outputs.repo` was already being set to `inputs.briefcase-url` earlier on in the Action if `Briefcase-Repo` wasn't specified in the PR Body; so, falling back here was already superfluous. The same is still true now for `steps.briefcase.outputs.repo`. As such, I just removed the fallback.

The Briefcase ref, however, has the extra step of deriving the ref from context; so, that still needs to be resolved here if the ref isn't being specified in the PR Body or as an input.

.github/actions/pr-body-parse/action.yml
rmartin16: I annotated the output with `extracted` to help provide some breadcrumbs later on. That is, when you actually reference this output, it is completely detached from this Action and is only associated with the `id` of the step that ran the action; for instance, `steps.briefcase-repo-override.outputs.extracted-value`. So, having `extracted` helps remind you that this value was parsed from an arbitrary source.

That said, is that useful information later on? maybe, maybe not, I guess. But it does make these variables even longer than they already are.

rmartin16:
Since I didn't explicitly solicit any feedback in my previous post :) what do you think about updating the Action to accept multiple regexes? Do you think that would improve the situation/implementation?

freakboy3742 (approved):

freakboy3742:
I guess it might be a little shorter, but I'm not sure it makes it shorter by enough to make the drop in functionality worth it. If we're doing this refactor, it's because we forsee a need to extract *other* variables from the PR body; limiting ourself to a fixed number of regexes seems like an odd choice to make under those circumstances.

The code that is here looks fine to me; I guess my question to you is whether you see it actually being useful in the longer term. I'm not 100% sure I do... but that said, you've written the PR, and we *might* use it... I've marked it as approved; but I'll leave it in your court as to whether we actually land it, or close it out as an interesting experiment but ultimately not worth it.

rmartin16:
ehh....we don't have a use for it now (or even a foreseeable one) and neither of us really like the refactor's outcome. we can come back to it if we need to.

PR #172, opened 2024-10-09
Enable allow-prereleases on all Python installs.

freakboy3742:
Enable the use of [`allow-prereleases`](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#allow-pre-releases) when setting up Python.

This allows us to use a bare version number for pre-release Python versions, rather than needing to add (and often remove at runtime) the `-dev` suffix. Given we're about to do a big push of adding 3.14, and removing the -dev suffix from  3.13, it seemed opportune to get this in place.

## PR Checklist:

### DIFF
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -223,7 +223,7 @@ jobs:
     - name: Build Linux System Project (Ubuntu, local)
       if: >
         startsWith(inputs.runner-os, 'ubuntu')
-        && inputs.python-version == "system"
+        && inputs.python-version == 'system'
         && contains(fromJSON('["", "Linux"]'), inputs.target-platform)
         && contains(fromJSON('["", "system"]'), inputs.target-format)
       working-directory: ${{ steps.create.outputs.project-path }}
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -111,10 +111,14 @@ jobs:
         path: ${{ steps.config.outputs.briefcase-data-dir }}
 
     - name: Set Up Python
-      if: ${{ inputs.python-version != 'system' }}
+      # On Linux, we can use the system python as-is. On macOS/Windows, install
+      # a Python that matches the version provided as a system default. This is
+      # to avoid issues with trying to install packages into homebrew's python
+      # without requiring sudo.
+      if: ${{ !startsWith(inputs.runner-os, 'ubuntu') || inputs.python-version != 'system' }}
       uses: actions/setup-python@v5.2.0
       with:
-        python-version: ${{ inputs.python-version }}
+        python-version: ${{ inputs.python-version == 'system' && steps.config.system-python-version || inputs.python-version }}
         allow-prereleases: true
         cache: pip
         cache-dependency-path: |
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -82,9 +82,6 @@ jobs:
         fi
         echo "cache-key=$(date +%Y-%m)|${CACHE_KEY}" | tee -a ${GITHUB_OUTPUT}
 
-        SYSTEM_PYTHON_VER=$(python3 -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")')
-        echo "system-python-version=${SYSTEM_PYTHON_VER}" | tee -a ${GITHUB_OUTPUT}
-
     - name: Checkout ${{ inputs.repository }}
       uses: actions/checkout@v4.2.0
       with:
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -114,7 +114,7 @@ jobs:
       if: ${{ !startsWith(inputs.runner-os, 'ubuntu') || inputs.python-version != 'system' }}
       uses: actions/setup-python@v5.2.0
       with:
-        python-version: ${{ inputs.python-version }}
+        python-version: ${{ inputs.python-version || '3.x' }}
         allow-prereleases: true
         cache: pip
         cache-dependency-path: |
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -10,7 +10,7 @@ on:
     inputs:
       python-version:
         description: "Python version to use; defaults to latest Python release."
-        default: "3.X"
+        default: "3.x"
         type: string
       runner-os:
         description: "The OS to use to build the App; must be a fully qualified GitHub runner OS, e.g. ubuntu-latest."
@@ -116,6 +116,7 @@ jobs:
       uses: actions/setup-python@v5.2.0
       with:
         python-version: ${{ inputs.python-version }}
+        allow-prereleases: true
         cache: pip
         cache-dependency-path: |
           **/setup.cfg
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -111,8 +111,7 @@ jobs:
         path: ${{ steps.config.outputs.briefcase-data-dir }}
 
     - name: Set Up Python
-      # Linux System apps can only use the system Python to run the app
-      if: ${{ !startsWith(inputs.runner-os, 'ubuntu') || inputs.python-version != 'system' }}
+      if: ${{ inputs.python-version != 'system' }}
       uses: actions/setup-python@v5.2.0
       with:
         python-version: ${{ inputs.python-version }}
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -118,7 +118,7 @@ jobs:
       if: ${{ !startsWith(inputs.runner-os, 'ubuntu') || inputs.python-version != 'system' }}
       uses: actions/setup-python@v5.2.0
       with:
-        python-version: ${{ inputs.python-version == 'system' && steps.config.system-python-version || inputs.python-version }}
+        python-version: ${{ inputs.python-version == 'system' && steps.config.outputs.system-python-version || inputs.python-version }}
         allow-prereleases: true
         cache: pip
         cache-dependency-path: |
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -111,14 +111,13 @@ jobs:
         path: ${{ steps.config.outputs.briefcase-data-dir }}
 
     - name: Set Up Python
-      # On Linux, we can use the system python as-is. On macOS/Windows, install
-      # a Python that matches the version provided as a system default. This is
-      # to avoid issues with trying to install packages into homebrew's python
-      # without requiring sudo.
+      # On Linux, accept "system" as a proxy for "don't install Python"; this will
+      # fall back to the system-provided Python package. Providing "system" as a value
+      # on macOS or Windows will cause an error.
       if: ${{ !startsWith(inputs.runner-os, 'ubuntu') || inputs.python-version != 'system' }}
       uses: actions/setup-python@v5.2.0
       with:
-        python-version: ${{ inputs.python-version == 'system' && steps.config.outputs.system-python-version || inputs.python-version }}
+        python-version: ${{ inputs.python-version }}
         allow-prereleases: true
         cache: pip
         cache-dependency-path: |
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -111,8 +111,8 @@ jobs:
         path: ${{ steps.config.outputs.briefcase-data-dir }}
 
     - name: Set Up Python
-      # Linux System apps requires python is System Python to run the app
-      if: ${{ !startsWith(inputs.runner-os, 'ubuntu') || !startsWith(inputs.python-version, steps.config.outputs.system-python-version) }}
+      # Linux System apps can only use the system Python to run the app
+      if: ${{ !startsWith(inputs.runner-os, 'ubuntu') || inputs.python-version != 'system' }}
       uses: actions/setup-python@v5.2.0
       with:
         python-version: ${{ inputs.python-version }}
@@ -223,7 +223,7 @@ jobs:
     - name: Build Linux System Project (Ubuntu, local)
       if: >
         startsWith(inputs.runner-os, 'ubuntu')
-        && startsWith(inputs.python-version, steps.config.outputs.system-python-version)
+        && inputs.python-version == "system"
         && contains(fromJSON('["", "Linux"]'), inputs.target-platform)
         && contains(fromJSON('["", "system"]'), inputs.target-format)
       working-directory: ${{ steps.create.outputs.project-path }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -291,7 +291,7 @@ jobs:
     needs: pre-commit
     uses: ./.github/workflows/app-build-verify.yml
     with:
-      python-version: "3.11"
+      python-version: "3.11"  # Explicitly test the python-version override
       repository: beeware/briefcase-template
       runner-os: ${{ matrix.runner-os }}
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
rmartin16, line 1: Few thoughts reviewing this workflow:
- All the testing of `app-build-verify.yml` explicitly specifies `python-version`; this seems unnecessary and should probably just use the default `3.x`
- For the Linux System builds, I think we should update `app-build-verify.yml` to allow the value `system` for `python-version`; that would signal `app-build-verify.yml` to skip `actions/setup-python` without the caller needing to explicitly specify the version of Python for the runner.

rmartin16 (approved):
👍🏼

.github/workflows/ci.yml
freakboy3742: > Few thoughts reviewing this workflow:
> 
> * All the testing of `app-build-verify.yml` explicitly specifies `python-version`; this seems unnecessary and should probably just use the default `3.x`

I must be looking at something else, because to my eye, all the tests of app-build-verify (in ci.yml) explicitly set a python version:
* test-verify-apps-briefcase - 3.10
* test-verify-apps-briefcase-template - 3.11
* test-verify-apps-android-templates - 3.11
* test-verify-apps-iOS-templates - 3.11
* test-verify-apps-linux-system-templates - 3.10
* test-verify-apps-appimage-templates - 3.11
* test-verify-apps-flatpak-templates - 3.11
* test-verify-apps-macOS-templates - 3.11
* test-verify-apps-web-templates - 3.11
* test-verify-apps-windows-templates - 3.11

> * For the Linux System builds, I think we should update `app-build-verify.yml` to allow the value `system` for `python-version`; that would signal `app-build-verify.yml` to skip `actions/setup-python` without the caller needing to explicitly specify the version of Python for the runner.

Agreed this would be a useful feature to add. 

To "yes-and" this - should we replace the existing system-python handling? The current handling has a bunch of "if the selected python version is the system python, don't install/do run this step" etc; should we replace that with "if the selected python version *is* `system`", and if the user happens to select a version of Python that happens to be the system python, then they get a "normal" python install?

rmartin16: > I must be looking at something else, because to my eye, all the tests of app-build-verify (in ci.yml) explicitly set a python version:

Right; I'm just saying I think we should _not_ set `python-version` in `ci.yml` and instead just let it fallback to the default (i.e. `3.x`) specified in `app-build-verify.yml`.

> To "yes-and" this - should we replace the existing system-python handling?

Yeah; this has all grown organically a bit and these simplifications definitely make sense to me.

freakboy3742: > Right; I'm just saying I think we should _not_ set `python-version` in `ci.yml` and instead just let it fallback to the default (i.e. `3.x`) specified in `app-build-verify.yml`.

Ah - agreed, except for keeping one as a token check of the "explicitly asking for a python version" case.
 
> > To "yes-and" this - should we replace the existing system-python handling?
> 
> Yeah; this has all grown organically a bit and these simplifications definitely make sense to me.

Ok - I've made these two changes - let me know what you think.

rmartin16 (changes requested):
I was initially confused reading some of the logic...but then I realized this is taking a bit more liberal view of "system" than what I was imagining.

When I think about "system Python", I am referring to the Python that is default and native to the environment. For Linux, this Python is governed by the distributors...but for macOS and Windows, this Python does not exist (or at least not meaningfully). Therefore, if we expand "system Python" to include macOS and Windows in CI, we're really just saying "whatever GitHub ships in the runner" is the "system Python". So, testing against a "system Python" on these platforms doesn't have any analogue in reality.

As another line of reasoning, this also creates a very different definition of "system" for non-Linux platforms. For Linux, we literally want to use the Python installed by the system. Elsewhere, though, we're saying we want a system Python...but really just install a version of GitHub's Python for us to use. Furthermore, we're depending on GitHub to not let this "system Python" be an EOL version at any point.

Philosophy aside...this implementation does have the advantage of simplifying the use of `app-build-verify` and `app-create-verify` insofar as allowing callers to just pass `system` for `python-version` regardless of the runner OS. If we take my imagined implementation, callers would need to conditionally set `python-version` to `system` only when testing native Linux packaging.

For instance, if we consider how Briefcase would need to use `app-build-verify`:
```yaml
verify-apps:
  name: Build app
  needs: [ package, package-automation, unit-tests ]
  uses: beeware/.github/.github/workflows/app-build-verify.yml@main
  with:
    python-version: ${{ matrix.python-version }}
    runner-os: ${{ matrix.runner-os }}
    framework: ${{ matrix.framework }}
  strategy:
    fail-fast: false
    matrix:
      framework: [ "toga", "pyside6", "pygame", "console" ]
      runner-os: [ "macos-12", "macos-14", "ubuntu-latest", "windows-latest" ]
      include:
        - runner-os: "ubuntu-latest"
          python-version: "system"
```

So, long-winded as usual...but personally, I think I lean towards a more limited definition of "system" where we're literally just talking about `/usr/bin/python3`. Therefore, the value of `system` for `python-version` would only mean "do not run `actions/setup-python`"....and this is really only appropriate for the Ubuntu runners. Thoughts?

freakboy3742:
> So, long-winded as usual...but personally, I think I lean towards a more limited definition of "system" where we're literally just talking about `/usr/bin/python3`. Therefore, the value of `system` for `python-version` would only mean "do not run `actions/setup-python`"....and this is really only appropriate for the Ubuntu runners. Thoughts?

That's where I started; the problem is that it fall apart on matrix builds because [you can't pass "system" into a macOS/Windows build in a matrix](https://github.com/beeware/.github/actions/runs/11303905027/job/31441632641). 

I guess the real question here is whether it makes more sense to raise an error, or fall back to a "recent version". I've used the fallback, using the python3 that is available in the CI platform; but there's an argument to be made that "3.x" might be a more predictable interpretation. I agree that it's a little weird in the intepretation of "system"; but it's a strict addition of behavior ("system" on macOS/Windows ~= "3.x"), and that interpretation doesn't *not* make sense given there's no other meaningful interpretation of system (unless you want to get into "Xcode ships with 3.9" and "Windows ships with a store shim that will install 3.x"... but a strict "3.x" makes more sense to me)

Having a fallback doesn't preclude the sort of usage you've described, either. You could still define a default python version of "3.12", and override the Python to "system" for a specific Linux build. Your specific example workflow needs one more detail, as the macOS/Windows cases won't define a value for python-version, but that's easy to add with either a default `includes:` value, or a ``${{ matrix.python-version || '3.12' }}` when defining the python version. 

The other detail worth flagging; app-create-verify *does* need to install Python, even on Linux, because otherwise it [tries to install dependencies into the system python, falls back to a user install and fails](https://github.com/beeware/.github/actions/runs/11303905027/job/31441632779). Open to suggestions to any

So - do we keep the "working fallback", or go back to "raise an error if you use 'system' on macOS/Windows", and modify the handful of matrix test cases to avoid the error?

rmartin16:
> So - do we keep the "working fallback", or go back to "raise an error if you use 'system' on macOS/Windows", and modify the handful of matrix test cases to avoid the error?

I mostly just want to avoid any behavior where passing `system` for `python-version` has a valid meaning for macOS and Windows; so, I think that should be a failure mode.

> Your specific example workflow needs one more detail, as the macOS/Windows cases won't define a value for python-version, but that's easy to add with either a default `includes:` value, or a `${{ matrix.python-version || '3.12' }}` when defining the python version.

In my example, I was intending the choice of Python version be deferred to `app-build-verify` since it should default to `3.x` in absence of a value.

> The other detail worth flagging; app-create-verify _does_ need to install Python, even on Linux, because otherwise it [tries to install dependencies into the system python, falls back to a user install and fails](https://github.com/beeware/.github/actions/runs/11303905027/job/31441632779). Open to suggestions to any fixes or workarounds on this one; but that's also the pre-existing behavior.

I think your solution is fine...but as with `app-build-verify`, I don't think `system` should have a meaning on macOS and Windows.


As an alternative, I suppose, we could expand a bit what we're trying to do. My initial thought was "let's embed this whole Python version pinning within the `beeware/.github` workflows" because the callers don't actually care. However, we could just take this a step further and use a value like `resolve` for `python-version`; in this way, the callers would then be explicitly deferring to the `beeware/.github` workflows to figure out the correct Python to run the workflow with.

freakboy3742:
> > So - do we keep the "working fallback", or go back to "raise an error if you use 'system' on macOS/Windows", and modify the handful of matrix test cases to avoid the error?
> 
> I mostly just want to avoid any behavior where passing `system` for `python-version` has a valid meaning for macOS and Windows; so, I think that should be a failure mode.

👍 

I can live with that, especially considering...
 
> > Your specific example workflow needs one more detail, as the macOS/Windows cases won't define a value for python-version, but that's easy to add with either a default `includes:` value, or a `${{ matrix.python-version || '3.12' }}` when defining the python version.
> 
> In my example, I was intending the choice of Python version be deferred to `app-build-verify` since it should default to `3.x` in absence of a value.

... this. I forgot that the workflow had a fallback value at the input level.
 
> > The other detail worth flagging; app-create-verify _does_ need to install Python, even on Linux, because otherwise it [tries to install dependencies into the system python, falls back to a user install and fails](https://github.com/beeware/.github/actions/runs/11303905027/job/31441632779). Open to suggestions to any fixes or workarounds on this one; but that's also the pre-existing behavior.
> 
> I think your solution is fine...but as with `app-build-verify`, I don't think `system` should have a meaning on macOS and Windows.

I've modified the implementation to take this approach.
 
> As an alternative, I suppose, we could expand a bit what we're trying to do. My initial thought was "let's embed this whole Python version pinning within the `beeware/.github` workflows" because the callers don't actually care. However, we could just take this a step further and use a value like `resolve` for `python-version`; in this way, the callers would then be explicitly deferring to the `beeware/.github` workflows to figure out the correct Python to run the workflow with.

I think I'd rather stick to the side of "explicit is better than implicit" - I'm not sure we reliably have enough context on the `.github` side of things to pick the "right" Python version reliably, unless we just fall back to "3.x"... which will be the default value if unspecified anyway. All we'd be encoding is an explicit value which is the implicit value of 3.x, except for linux, which is "system", which I don't think is enough benefit to be worth it.

rmartin16 (changes requested):
Some final details to work out.

.github/workflows/ci.yml
```diff
@@ -285,13 +287,17 @@ jobs:
       matrix:
         framework: [ toga ]
         runner-os: [ macos-latest, windows-latest, ubuntu-22.04 ]
+        includes:
+          # Ubuntu apps need to run on the system Python
+          - runner-os: ubuntu-22.04
+            python-version: "system"
```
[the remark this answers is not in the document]
freakboy3742: Good catch. And yeah - not sure there's anything to be done here, other than removing the default value in the workflow... which isn't that much better as an outcome, really.

.github/workflows/app-build-verify.yml
```diff
@@ -111,11 +111,14 @@ jobs:
         path: ${{ steps.config.outputs.briefcase-data-dir }}
 
     - name: Set Up Python
-      # Linux System apps requires python is System Python to run the app
-      if: ${{ !startsWith(inputs.runner-os, 'ubuntu') || !startsWith(inputs.python-version, steps.config.outputs.system-python-version) }}
+      # On Linux, accept "system" as a proxy for "don't install Python"; this will
+      # fall back to the system-provided Python package. Providing "system" as a value
+      # on macOS or Windows will cause an error.
+      if: ${{ !startsWith(inputs.runner-os, 'ubuntu') || inputs.python-version != 'system' }}
```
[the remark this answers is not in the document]
freakboy3742: I think my preference here is to update the downstream usage. Having some Python versions mysteriously not install is a lot less obvious than an explicit "use the system version".

.github/workflows/app-build-verify.yml
```diff
@@ -222,7 +225,7 @@ jobs:
     - name: Build Linux System Project (Ubuntu, local)
       if: >
         startsWith(inputs.runner-os, 'ubuntu')
-        && startsWith(inputs.python-version, steps.config.outputs.system-python-version)
+        && inputs.python-version == 'system'
         && contains(fromJSON('["", "Linux"]'), inputs.target-platform)
         && contains(fromJSON('["", "system"]'), inputs.target-format)
```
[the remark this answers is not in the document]
freakboy3742: Agreed. :-)

.github/workflows/app-build-verify.yml
```diff
@@ -111,11 +108,14 @@ jobs:
...
-      # Linux System apps requires python is System Python to run the app
-      if: ${{ !startsWith(inputs.runner-os, 'ubuntu') || !startsWith(inputs.python-version, steps.config.outputs.system-python-version) }}
+      # On Linux, accept "system" as a proxy for "don't install Python"; this will
+      # fall back to the system-provided Python package. Providing "system" as a value
+      # on macOS or Windows will cause an error.
+      if: ${{ !startsWith(inputs.runner-os, 'ubuntu') || inputs.python-version != 'system' }}
       uses: actions/setup-python@v5.2.0
       with:
-        python-version: ${{ inputs.python-version }}
+        python-version: ${{ inputs.python-version || '3.x' }}
```
freakboy3742, line 117: By way of explanation: It appears "with: ''" is passed through as a literal empty string, rather than falling back to a default, so we need to provide the default as a fallback here, as well (same for the create-verify analog).

rmartin16 (approved):
Looks good. Feel free to add me to the downstream PRs.

freakboy3742:
> Looks good. Feel free to add me to the downstream PRs.

Will do... however, I don't think there's as many as I originally thought. Briefcase and the linux system template, obviously... but the other Linux templates don't need it... what else am I missing?

rmartin16:
I think the explicit use of `ubuntu-22.04` should be a good signal:
```
~/github/beeware took 6s 
❯ grep -rF ubuntu-22.04 | grep 'yml\|yaml'
briefcase/.readthedocs.yaml:  os: ubuntu-22.04
briefcase/.github/workflows/ci.yml:        runner-os: [ "macos-12", "macos-14", "ubuntu-22.04", "windows-latest" ]
briefcase/.github/workflows/ci.yml:      # Ubuntu version used to run Linux tests. We use a fixed ubuntu-22.04
briefcase/.github/workflows/ci.yml:        runner-os: [ "macos-12", "macos-14", "ubuntu-22.04", "windows-latest" ]

toga-chart/.readthedocs.yaml:  os: ubuntu-22.04

toga/.readthedocs.yaml:  os: ubuntu-22.04
toga/.github/workflows/ci.yml:          runs-on: "ubuntu-22.04"
toga/.github/workflows/ci.yml:          runs-on: "ubuntu-22.04"
toga/.github/workflows/ci.yml:          runs-on: "ubuntu-22.04"

.github/.github/workflows/ci.yml:        runner-os: [ macos-latest, windows-latest, ubuntu-22.04 ]
.github/.github/workflows/ci.yml:        runner-os: [ macos-latest, windows-latest, ubuntu-22.04 ]
.github/.github/workflows/ci.yml:      runner-os: ubuntu-22.04

rubicon-objc/.readthedocs.yml:  os: ubuntu-22.04

beeware/.readthedocs.yaml:  os: ubuntu-22.04

briefcase-template/.github/workflows/ci.yml:        runner-os: [ "macos-latest", "ubuntu-22.04", "windows-latest" ]
briefcase-template/.github/workflows/ci.yml:      # ubuntu-22.04, which is needed to test local Debian packages. We use
briefcase-template/.github/workflows/ci.yml:      # ubuntu-22.04 explicitly rather than ubuntu-latest because when
briefcase-template/.github/workflows/ci.yml:        runner-os: [ "macos-latest", "ubuntu-22.04", "windows-latest" ]
```

So, looks like just `briefcase-template`.

freakboy3742:
> So, looks like just `briefcase-template`.

Good catch - I've pushed a PR for that one as well.

The others all look like either explicit RTD configs, explicit configs in *this* repo, PR's I've already submitted, or Toga configs that need to be explicitly 22.04.

PR #186, opened 2024-11-18
Exclude pyproject.toml from dependency version bumps

### LINKED ISSUE #185: setuptools updates processed twice
### Describe the bug

In the Nov 17 weekly updates, Setuptools was bumped from 75.3.0 to 75.5.0. This definition exists as a build dependency in pyproject.toml, so it was picked up by the Brutus update bot.

However, the same update was picked up by Dependabot, leading to pairs of overlapping updates (e.g., beeware/briefcase#2061 and beeware/briefcase#2063). Once the Brutus update bot PR is landed, dependabot rebases, and cleans up; but the inverse isn't true.

This appears to be the result of a recent update to dependabot. As of [v0.286.0](https://github.com/dependabot/dependabot-core/releases/tag/v0.286.0) (released 3 days ago), dependabot [has support for pyproject.toml buildsystem.requires](dependabot/dependabot-core/pull/10899). 

### Steps to reproduce

Not sure there's a clear way to reproduce this, other than putting a package into pyproject.toml as a dependency, then upgrading it, then re-running both bots.

### Expected behavior

setuptools (or any other pyproject.toml build dependency) should only be picked up once.

### Screenshots

_No response_

### Environment

- Operating System: Github Actions

### Logs

```

```

### Additional context

We should keep an eye out for dependency group support in dependabot; that would remove the need for the autoupdate bot entirely.

rmartin16:
## Changes
- Dependabot now handles pyproject.toml; only bump tox.ini
- Fixes #185

## PR Checklist:

### DIFF
--- a/.github/workflows/dep-version-bump.yml
+++ b/.github/workflows/dep-version-bump.yml
@@ -14,6 +14,10 @@ on:
         description: "Whitespace-delimited list of directories containing pyproject.toml and tox.ini files; defaults to repo's base directory."
         default: ""
         type: string
+      filenames:
+        description: "Whitespace-delimited list of filenames to evaluate for dependency version bumps; defaults to only tox.ini."
+        default: "tox.ini"
+        type: string
       create-changenote:
         description: "Defaults 'true' to create a misc changenote in the './changes' directory."
         default: true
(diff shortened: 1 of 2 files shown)


freakboy3742 (approved):
👍 Thanks for jumping on this so quickly.

PR #255, opened 2025-10-25
no differentiator when subdirectory is root directory

timrid:
In https://github.com/beeware/briefcase/pull/2173 I was advised to use the [`artifact-basename` trick that Toga uses](https://github.com/beeware/briefcase/pull/2173#pullrequestreview-3182936273) in briefcase too. 

I tried packaging all packages in one job like this in [`ci.yml`](https://github.com/timrid/briefcase/blob/df74eba6cee61c5bba0bd6b12109a3b044749324/.github/workflows/ci.yml#L39-L57):

```yml
  package:
    name: Package Briefcase
    permissions:
      id-token: write
      contents: read
      attestations: write
    strategy:
      matrix:
        subdir:
        - "."
        - "debugger"
        - "automation"
    uses: beeware/.github/.github/workflows/python-package-create.yml@main
    with:
      build-subdirectory: ${{ matrix.subdir }}
      attest: ${{ inputs.attest-package }}
```

But then the original `Packages-briefcase` artifact was renamed to `Packages-briefcase-.`. An that generates an error [in `app-create-verify.yml`](
https://github.com/beeware/.github/blob/56040316f82227519c03eb0b8880f2dea56e0326/.github/workflows/app-create-verify.yml#L100), that enforces the name `Packages-briefcase`.

So this PR skips the differentiator when passing the root (`.`) as`build-subdirectory`.

(This isn't quite right, since `.` isn't actually a subdirectory, but I didn't see any way to optionally skip the `build-subdirectory` parameter in `ci.yml`.)

## PR Checklist:

### DIFF
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -147,7 +147,7 @@ jobs:
     - name: Get Briefcase Packages
       # Briefcase will build and package itself in a previous step in its CI
       if: endsWith(inputs.repository, 'briefcase')
-      uses: actions/download-artifact@v5.0.0
+      uses: actions/download-artifact@v6.0.0
       with:
         # artifact name is derived in python-package-create.yml
         pattern: "Packages-briefcase*"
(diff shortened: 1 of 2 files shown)


timrid:
> Regarding the specific change in this PR - I think you could likely get the same effect by specifying "" as the subdirectory. However, that's even less intuitive than ".", so making this change makes sense to me.

How could I have missed that? It works like a charm. Then this PR here becomes redundant.

PR #260, opened 2025-11-01
Run pre-commit with Python 3.13 by default

rmartin16:
## Changes
- GitHub runners are turning over to Python 3.14 as default but docformatter cannot run on 3.14
- This will default pre-commit to run on 3.13 until resolved

## PR Checklist:

### DIFF
--- a/.github/workflows/pre-commit-run.yml
+++ b/.github/workflows/pre-commit-run.yml
@@ -9,7 +9,7 @@ on:
     inputs:
       python-version:
         description: "Python version to use; defaults to latest Python release."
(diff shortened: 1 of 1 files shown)


NMertsch:
References:
* Issue in docformatter: https://github.com/PyCQA/docformatter/issues/327
* Caused by issue in untokenize: https://github.com/myint/untokenize/issues/4

freakboy3742 (approved):
👍 Thanks - good to catch this ahead of Sunday's Dependabot run. Have to imagine it would cause some hilarity...

PR #271, opened 2025-12-03
Free up disk space on Linux runner

mhsmith:
Should fix this failure: https://github.com/beeware/briefcase/actions/runs/19875943952/job/56963552104

Android emulator failures started happening around the same time in the CPython repository, e.g. [this run](https://github.com/python/cpython/actions/runs/19897211275/job/57030696234). The error message is completely different, but emulator error reporting has always been poor, so it's possible it has the same cause.

Toga appears unaffected so far.

## PR Checklist:

### DIFF
--- a/.github/workflows/app-build-verify.yml
+++ b/.github/workflows/app-build-verify.yml
@@ -114,7 +114,7 @@ jobs:
 
         # Free up disk space
         docker system prune -a -f
-        sudo rm -rf $ANDROID_HOME/ndk/{26,28}.* /opt/hostedtoolcache/CodeQL \
+        sudo rm -rf $ANDROID_HOME/ndk /opt/hostedtoolcache/CodeQL \
           /usr/local/lib/node_modules /usr/local/share/chromium \
           /usr/local/share/powershell
         df -h


mhsmith:
pybind11 started hitting the same errors as CPython 2 weeks ago: https://github.com/pybind/pybind11/pull/5914.

mhsmith:
I copied this script from the cibuildwheel CI, which removes all NDK versions except the one it uses. But Briefcase doesn't use the NDK at all anymore, so I've changed it to remove all versions.

freakboy3742 (approved):
This unblocks the immediate problem, but I don't love this. While we might not need the NDK (or node, chromium, CodeQL) right now, this seems like the sort of thing that might bite us in the future when we suddenly *do* need one of these tools, and the GitHub docs say they should be available, and we forget that our CI deletes them.

If it's disk space that is the issue. I wonder if doing a more aggressive cleanup after each individual build might be a better way to go (i.e., build the system app, then delete the build/dist folders; build the flatpak app, then delete the folders and purge the flatpak archive etc). 

I'll merge this as an immediate fix; but I'll have a poke around to see if I can make a "cleanup" approach work.

PR #303, opened 2026-02-06
Update bug report issue template.

kattni:
Includes the following:

- Adds required field for minimum code example, formatted as code, with applicable description.
- Logs input now renders as code input automatically, and the description now reflects that.
- The Environment section updated with clearer description and value.
- Updates the rest of the descriptions to have consistent punctuation.

## PR Checklist:
 
### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -26,7 +26,7 @@ body:
   - type: textarea
     id: reproduce
     attributes:
-      label: The steps to reproduce
+      label: How to reproduce the bug
       description: Provide a step-by-step explanation of how to reproduce the bug.
       value: |
         1. Go to '...'


mhsmith:
Please let's not add another required field. There are too many already, and in most of the issues I create, I end up having to put placeholders in them all and then edit the issue after creation to delete them.

I would suggest instead changing the titles or descriptions of the existing fields to list the kinds of things that should go into them.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -7,68 +7,76 @@ body:
...
-    attributes:
-      label: Logs
-      description: Paste or upload relevant log files. Briefcase will create a logfile when invoked with the ``--log`` flag.
+      label: Screenshots
+      description: Please add screenshots of the bug happening, if applicable.
       value: |
-        ```
-
-        ```
+        <!-- If the bug is visual, please include screenshots of the bug happening. -->
```
mhsmith: I don't think there's much point in including placeholder text if it's basically the same as the description.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -7,68 +7,76 @@ body:
...
     validations:
       required: true
 
   - type: textarea
-    id: screenshots
     attributes:
-      label: Screenshots
-      description: Please add screenshots, if applicable.
+      label: Logs
+      description: Paste or upload relevant log files. Briefcase will create a logfile when invoked with the ``--log`` flag. This will be automatically formated into code, so there is no need for backticks.
```
mhsmith:
```suggestion
      description: Paste or upload relevant log files. Briefcase will create a logfile when it has an error, or when invoked with the ``--log`` flag. This will be automatically formated into code, so there is no need for backticks.
```

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -7,68 +7,76 @@ body:
...
+      render: shell
```
mhsmith: Will this `render` setting interfere with uploading a file into the field?

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -7,68 +7,76 @@ body:
...
     validations:
       required: true
 
   - type: textarea
-    id: expected
+    id: example
     attributes:
-      label: Expected behavior
-      description: A clear and concise description of what you expected to happen
+      label: Minimum example code
```
mhsmith: I agree we want example code if it's applicable, but I don't think we should make it a separate required field, because it's not always applicable. For example, if someone has a usage problem with Briefcase, there isn't any code involved.

I'd suggest something like this:
* Describe the bug
  * (In placeholder comment) Include all of the following, where applicable:
    * A step-by-step explanation of what you did and what happened.
    * What you expected to happen.
    * The smallest possible code example that is needed to reproduce the bug.
 * Environment
   * Separate required field, like before.

.github/ISSUE_TEMPLATE/bug_report.yml
kattni: Confirmed, yes, it does interfere. I can revert it. Alternatively, we indicate you can include a file in the "Additional context" section, if you aren't pasting something in.

PR #306, opened 2026-02-14
Adds documentation issue template.

### LINKED ISSUE #305: Add documentation issue template
The org-level issue templates in this repo provide Bug report and Feature request options. There is no template for documentation issues (missing, incorrect, or unclear docs).

Documentation issues don't fit either existing template. A bug report asks for reproduction steps, expected behavior, environment, and logs, none of which apply when a reference page is missing a section. A feature request asks for a problem, a proposed solution, and alternatives considered, which doesn't match "this documented thing exists but isn't on the page." Forcing docs issues into one of these templates means reporters either leave required fields empty or write "N/A" through them.

`config.yml` has `blank_issues_enabled: false`, so the web UI offers no escape hatch, reporters must pick bug or feature. I ran into this filing beeware/briefcase#2681 (a missing section in the plugins reference page). Using `gh issue create` from the CLI worked, but I couldn't apply a `documentation` label because there's no template that provides one and the web UI wouldn't have offered a fitting path at all. Contributors who don't use the CLI have no way to file a docs issue without misusing one of the existing templates.

Adding a `documentation.yml` template would give reporters a structured way to file docs issues with the relevant fields (affected page, what's missing/wrong, suggested improvement) and auto-apply the `documentation` label.

filiplajszczak:
Resolves #305

## PR Checklist:

### DIFF
--- a/.github/ISSUE_TEMPLATE/documentation.yml
+++ b/.github/ISSUE_TEMPLATE/documentation.yml
@@ -1,0 +1,41 @@
+name: 📝 Documentation
+description: Report missing, incorrect, or unclear documentation
+labels: ["documentation"]
+
+body:
+
+  - type: dropdown
+    id: kind
+    attributes:
+      label: What kind of documentation issue is this?
+      options:
+        - Missing documentation
+        - Incorrect documentation
+        - Unclear or confusing documentation
+    validations:
+      required: true
+
+  - type: input
+    id: page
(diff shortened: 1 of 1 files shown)


kattni (approved):
Thanks for adding this! It's a great place to start. I imagine we'll want to adapt it over time, but I'm quite happy with this as an initial template.

PR #356, opened 2026-05-19
Install workflow deps from pyproject.toml dependency groups

### LINKED ISSUE #355: Update app-create-verify.yml to install deps from pyproject.toml
### What is the problem or limitation you are having?

The `app-create-verify.yml` workflow has a hardcoded reference to `requirements.txt`:

https://github.com/beeware/.github/blob/main/.github/workflows/app-create-verify.yml#L92

```yaml
run: python -m pip install -Ur ./briefcase-template/requirements.txt
```

This blocks beeware/briefcase-template#236 (migrating that repo from `requirements.txt` to `pyproject.toml`), since deleting the file would break this workflow for all consumers.

### Describe the solution you'd like

Replace the hardcoded requirements.txt install with a dependency group read from pyproject.toml, and add a pip upgrade step (required for `--dependency-group` support):

```yaml
- name: Update pip
  run: python -m pip install -U pip

- name: Install Dependencies
  run: python -m pip install --dependency-group test ./briefcase-template
```

### Describe alternatives you've considered

N/A

### Additional context

Related to #295 (broader migration of this repo to `pyproject.toml`) but distinct. That issue covers this repo's own dependencies, while this workflow separately hardcodes a path into a consumer repo's `requirements.txt`.

Blocks beeware/briefcase-template#236.

gwdio:
Replace the hardcoded requirements.txt install with --group reads from briefcase-template's pyproject.toml, unblocking the migration in beeware/briefcase-template#236.

Update `app-create-verify.yml` to install briefcase-template's workflow dependencies via PEP 735 dependency groups (`--group`) instead of `requirements.txt`. Adds a pip upgrade step since `--group` requires pip ≥ 25.1.

Unblocks beeware/briefcase-template#236, which migrates that repo to `pyproject.toml`. The workflow currently hardcodes `pip install -Ur ./briefcase-template/requirements.txt`, so deleting the file in the consumer repo would break this workflow for all consumers. Once this lands, the temporary `requirements.txt` stub in briefcase-template can be removed.

Fixes #355
Refs beeware/briefcase-template#236

## PR Checklist:

- [x] I will abide by the BeeWare Code of Conduct
- [x] I have read and have followed the **CONTRIBUTING.md** file
- [x] This PR was generated or assisted using an AI tool

Assisted-by: Claude Opus 4.7

### DIFF
--- a/.github/workflows/app-create-verify.yml
+++ b/.github/workflows/app-create-verify.yml
@@ -91,7 +91,7 @@ jobs:
       run: python -m pip install -U pip
 
     - name: Install Dependencies
-      run: python -m pip install --group test --group dev ./briefcase-template
+      run: python -m pip install --group ./briefcase-template:test --group ./briefcase-template:dev
 
     - name: Get Briefcase Package
       # Briefcase will build and package itself in a previous step in its CI
--- a/.github/workflows/app-create-verify.yml
+++ b/.github/workflows/app-create-verify.yml
@@ -87,9 +87,11 @@ jobs:
           **/pyproject.toml
           .pre-commit-config.yaml
 
+    - name: Update pip
+      run: python -m pip install -U pip
+
     - name: Install Dependencies
-      # must install before Briefcase below since requirements.txt contains an entry for Briefcase
-      run: python -m pip install -Ur ./briefcase-template/requirements.txt
+      run: python -m pip install --group test --group dev ./briefcase-template
 
     - name: Get Briefcase Package
       # Briefcase will build and package itself in a previous step in its CI
--- a/.github/workflows/app-create-verify.yml
+++ b/.github/workflows/app-create-verify.yml
@@ -91,7 +91,7 @@ jobs:
       run: python -m pip install -U pip
 
     - name: Install Dependencies
-      run: python -m pip install --group ./briefcase-template:test --group ./briefcase-template:dev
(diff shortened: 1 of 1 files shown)


phildini (approved):
@freakboy3742 I have reviewed this and think it's good, and have considered the implications, but I think it needs at least a glance from you because it affects the full downstream pipelines.

.github/workflows/app-create-verify.yml
```diff
@@ -87,9 +87,11 @@ jobs:
           **/pyproject.toml
           .pre-commit-config.yaml
 
+    - name: Update pip
+      run: python -m pip install -U pip
+
     - name: Install Dependencies
-      # must install before Briefcase below since requirements.txt contains an entry for Briefcase
-      run: python -m pip install -Ur ./briefcase-template/requirements.txt
+      run: python -m pip install --group './briefcase-template/pyproject.toml:test' --group './briefcase-template/pyproject.toml:dev'
```
freakboy3742, line 94: TIL about the filename:group syntax for `--group`.

I think we might need to revisit the groupings in briefcase-template to make them a little more focussed. I think the set we need is:
1. pre-commit (to run the pre-commit step on briefcase-template)
2. tox (to run the unit-test step on briefcase-template)
3. briefcase (as an explicit target for *just* briefcase, which is what app-create and app-build should need)
4. test (the things needed to run the actual test suite - pytest, black, etc, including a link to the briefcase target)
5. dev (for local development, linking essentially all of the above into a single target)

PR #378, opened 2026-07-13
Audit permissions and add zizmor configuration

freakboy3742:
Audits the permission allocations for tasks, and adds a zizmor validation step.

The Zizmor rules being corrected are:
1. (unpinned-uses) Adding hashes for actions. This prevents an attack by pushing a malicious update (and tag change) to an action repository
2. (template-injection) Using environment variables instead of direct references to inputs. `${{ }}` is interpreted as a direct templated value, so a malicious PR could construct inputs that would exfiltrate credentials, etc.
3. (excessive-permissions) Adding permissions blocks everywhere
4. (secrets-inherit) Explicitly request permissions, rather than inheriting
5. (artipacked) Github credentials are cached on checkout, which potentially gives permission for repo-modifying actions. The key side effect of this is that we no longer push Docker images to the GHCR repository. This may slow down CI passes.
6. (bot-conditions) using `github.event.pusher.name` rather than `github.actor`, as the latter can be spoofed by a commit.
7. (dependabot-cooldown) Add a cooldown on dependabot so that problems with new releases (including malicious releases created by an attacker) can be identified.

See inline for some other notes.

## PR Checklist:

- [x] I will abide by the BeeWare Code of Conduct
- [x] I have read and have followed the **CONTRIBUTING.md** file
- [x] This PR was generated or assisted using an AI tool
Assisted-by: Claude Opus 4.8

### DIFF
--- a/.github/workflows/coc-update.yml
+++ b/.github/workflows/coc-update.yml
@@ -20,52 +20,68 @@ jobs:
       commit_hash: ${{ steps.commit_hash.outputs.commit_hash }}
     steps:
     - name: Checkout this repository
-      uses: actions/checkout@v7
+      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+      with:
+        persist-credentials: false
     - name: Get latest commit hash
       id: commit_hash
       run: |
         commit_hash=$(git rev-parse --short=16 HEAD)
-        echo "commit_hash=$commit_hash" >> $GITHUB_OUTPUT
+        echo "commit_hash=$commit_hash" >> "$GITHUB_OUTPUT"
     - name: Get list of repositories
       id: repo_list
       env:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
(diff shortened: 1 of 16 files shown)


.github/workflows/coc-update.yml
```diff
@@ -20,52 +20,68 @@ jobs:
...
       id: repo_list
       env:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: |
-        repo_list=$(gh repo list $GITHUB_REPOSITORY_OWNER --no-archived --json nameWithOwner --jq ".[].nameWithOwner" -L 500)
-        jq_repo_list=$(jq -c -n '$ARGS.positional' --args ${repo_list[@]})
-        echo repo_list="$jq_repo_list" >> $GITHUB_OUTPUT
+        repo_list=$(gh repo list "$GITHUB_REPOSITORY_OWNER" --no-archived --json nameWithOwner --jq ".[].nameWithOwner" -L 500 \
+          | jq -R -s -c 'split("\n") | map(select(length > 0))')
+        echo "repo_list=$repo_list" >> "$GITHUB_OUTPUT"
```
freakboy3742, line 38: In the old approach, `repo_list` was a single string, not an array; while it worked, it was mostly accidental. The actual splitting was occurring because of shell handling, which would have been sensitive to special characters (an unlikely situation, but an avoidable one regardless).

